Issue #83, opened 2025-10-03
asusctl installs broken udev rule


evertvorster:
Hi there!

asusctl on arch linux installs /usr/lib/udev/rules.d/99-asusd.rules. 

When asus-nb-wmi is loaded, it wants to restart the asusd.service. 

I get this in my journal on every boot:
```
Oct 03 06:46:20 Evert.Scar (udev-worker)[728]: asus-nb-wmi: Failed to find and pin callout binary "/usr/lib/udev/systemctl": No such file or directory
Oct 03 06:46:20 Evert.Scar (udev-worker)[728]: asus-nb-wmi: Failed to execute 'systemctl restart asusd.service', ignoring: No such file or directory
```

So, yes, there is no systemctl in /usr/lib/udev:
```
evert@Evert /u/l/u/rules.d> which systemctl
/usr/bin/systemctl
```

evertvorster:
Yeah, so udev RUN _must_ supply an absolute path for executables:
```
RUN{type}
Specify a program to be executed after processing of all the rules for the event. With "+=", this invocation is added to the list, and with "=" or ":=", it replaces any previous contents of the list. Please note that both "program" and "builtin" types described below share a common list, so clearing the list with ":=" and "=" affects both types.

type may be:

"program"
Execute an external program specified as the assigned value. If no absolute path is given, the program is expected to live in /usr/lib/udev; otherwise, the absolute path must be specified.

This is the default if no type is specified.
```
Source:
https://www.freedesktop.org/software/systemd/man/latest/udev.html?utm_source=chatgpt.com

PR #84, opened 2025-10-05
Update asusd.rules to have absolute path

evertvorster:
Added the absolute path for systemctl. 
No modern distrobution installs it anywhere else.

We might want to revisit this with a "--no-block" argument that will speed up startup times a little, as udev is not blocked on this process completing. Currently left it out so that this is the smallest possible change to fix the issue of the warning in the journal.


evertvorster:
fixes #83

PR #95, opened 2025-11-02
Allow setting default profile for ac and battery

Matszwe02:
This PR adds options:

```sh
asusctl profile -a Performance
asusctl profile -b Quiet
```

which override the default `Quiet` and `Performance` profiles on battery and ac respectfully


DBMTC:
Hi and thanks for this!

Can you please send this to gitlab.com/asus-linux/asusctl please?

PR #111, opened 2026-05-04
Add Ubuntu 22.04 build support, X11 GUI path, and uhid-virt patch

wing-kit:
- Document Jammy installs; add rust-toolchain.toml (1.82) and extra scripts
- GitLab CI: Ubuntu 22.04 release build; format uses cargo +nightly fmt
- Patch crates.io uhid-virt for uhidrs-sys 1.0.4 constant names
- rog-control-center: friendlier Slint loop failure hint; README X11 note
- One-liner installer for X11-only Ubuntu 22.04 users


NeroReflex:
Ubuntu 22 is too old to support, and I will not add X11 support. Perhaps I can see if I can improve uhid, what's the point of the patchset?

wing-kit:
To be honest I simply asked Cursor to make it work with my Ubuntu 22.04. And it just worked.

scardracs:
@wing-kit my comment will probably be harsh but you can't expect someone with a minimum of brain to merge a thing like that.

1. You, in primis, don't even know how the code you posted works: it's a huge amount of codeslop you can't even maintain in the future, forcing the maintainers to it pick up.
2. Maintaining something as old as Ubuntu 22.04 is not a solution and it will not be:
    - It's old.
    - Ubuntu is not officially maintained.
    - X11 is almost dead and should remain as is.
3. You are going to add lincenses here and there without even checking the legal aspects of it. This is WRONG, really REALLY wrong.
4. You have pushed a single commit for modifying 17 different files, without a minimum of check. Do you even know how commit works or is a IA slop too?

I'm not saying not to use a IA, I'm the first who uses it for my patches because I can't humanly check every single use case or remember everything, but codeslop is WRONG.

NeroReflex:
Exactly the point. The reason I asked about uhid is because ot might be the only thing on this PR that is worth looking into.

Everything else don't belong here. Development in general is based on the principle that who submit the PR has planned what would have been done, has spent time creating it, researching it etc.... Because when they are done this becomes code I am in charge of.

So when a PR is sent the effort of the user finishes and mine starts. If the original effort was giving vague instructions to an llm and not care to understand what it did, how can I possibly maintain this?

Beside, honestly, just stop using ubuntu 22.04. It's ancient. Stop using x11: it doesn't belongs in a laptop in 2026. If some other program is needed that o ly runs in ubuntu 22.04 there is a much better use of llm: ask it to make that software work in a ubuntu 22.04 container.... Or if it's open software what to do about it.

scardracs:
I know that's late but I'm sorry for my answer. I didn't mean to be mean in any way and I'm sorry if I made someone uncomfortable with what I said.

PR #126, opened 2026-06-17
add metainfo for rog-control-center

Owen-sz:
This brings the metainfo file from the Terra package into upstream


NeroReflex:
Thanks for this! Can you please add Chore: at the beginning of commit messages?

NeroReflex:
Actually nvm. I am going to squash this

PR #127, opened 2026-06-20
Feat: add aura support for FA608UH

Igilq:
Added support for FA608UH, which is missing. after manually testing all modes, only those that i added to file work


NeroReflex:
Hi: can you rename this commit to:

Feat: add aura support for FA608UH

Thanks!

PR #128, opened 2026-06-20
chore: create a TODO file with all the TODOs in the whole project

scardracs:
Since we are not using Gitlab anymore let's remove gitlab-related code.
For issues we need to set it up from the settings page and gitlab-ci.yml has been substituted by .github folder


NeroReflex:
@pastaq Can you change the issue template with the one that is being removed in here please?

PR #133, opened 2026-06-22
Fix: bidirectional Quiet<->LowPower fallback in platform_profile setters

f4mrfaux:
## Summary

On hardware where the kernel ACPI platform_profile sysfs exposes only one of `Quiet` / `LowPower` for the same semantic profile (different ASUS DMI quirks expose different names — e.g. GA503QR exposes `quiet balanced performance`, never `low-power`), setting the absent variant fails with:

```
org.freedesktop.DBus.Error.NotSupported:
RogPlatform: platform_profile: (low-power) not supported
```

even when the equivalent IS in choices. The existing fallback in `set_platform_profile_on_battery` / `set_platform_profile_on_ac` only handled `Quiet -> LowPower`, and the bare `set_platform_profile` D-Bus property had no fallback at all.

This PR makes the alias bidirectional everywhere it matters:

- `set_platform_profile` (asusd/src/ctrl_platform.rs:469) — substitutes `Quiet <-> LowPower` when the requested variant isn't in `platform_profile_choices` but its semantic equivalent is.
- `set_platform_profile_on_battery` (~line 523) and `set_platform_profile_on_ac` (~line 562) — extended from one-direction to bidirectional, so the **canonical** name (the one the kernel exposes) is what gets persisted to `/etc/asusd/asusd.ron`.
- `select_power_profile_for_source` (~line 260) — extended to normalize both directions at apply-time, fixing stale-config rehydration after reboot or AC/BAT transitions.

## Background — relation to #648

The CLI-string-parsing portion of [GitLab issue #648](https://gitlab.com/asus-linux/asusctl/-/issues/648) was fixed by [GitLab MR !226](https://gitlab.com/asus-linux/asusctl/-/merge_requests/226) (commit `9366b0e`) which made `low-power` / `lowpower` / `LowPower` / `low_power` all parse to `PlatformProfile::LowPower`. That fix is necessary but not sufficient: the parsed enum then reaches `set_platform_profile`, which checks `choices.contains(&policy)` and rejects on hardware exposing only the kernel-renamed sibling name. This PR closes that follow-up gap.

## EPP / behavior preservation

EPP outcomes are unchanged. Both `Quiet` and `LowPower` already map to the same source in:
- `get_config_epp_for_throttle` (`asusd/src/ctrl_platform.rs:254-255`) — both → `profile_quiet_epp`
- `impl From<PlatformProfile> for CPUEPP` (`rog-platform/src/cpu.rs:209-210`) — both → `CPUEPP::Power`

So substitution does not affect what EPP gets written; only what name is stored in config and which sysfs string the kernel sees.

## Test plan — executed locally on patched daemon

Tested live on GA503QR (linux-g14 7.0.5, `/sys/firmware/acpi/platform_profile_choices = quiet balanced performance`) via a swap-in-place patched daemon (`systemctl stop asusd; systemd-run /target/release/asusd`):

| # | Command | Pre-patch | Post-patch | Result |
|---|---------|-----------|------------|--------|
| 1 | `asusctl profile set LowPower` | `NotSupported (low-power)` | kernel=`quiet`, active=`Quiet` | ✅ FIX |
| 2 | `asusctl profile set Quiet` | works | works | ✅ no regression |
| 3 | `asusctl profile set Balanced` | works | works | ✅ no regression |
| 4 | `asusctl profile set Performance` | works | works | ✅ no regression |
| 5 | `asusctl profile set NotARealProfile` | CLI parser rejects | CLI parser rejects | ✅ no regression |
| 6 | `asusctl profile set LowPower -b` | broken / wrong config | config persists `Quiet` (canonical), applied | ✅ FIX |
| 7 | `asusctl profile set LowPower -a` | broken / wrong config | config persists `Quiet` (canonical), applied | ✅ FIX |

Config persistence verified — `/etc/asusd/asusd.ron` after `set LowPower -b -a`:
```
platform_profile_on_battery: Quiet,
platform_profile_on_ac: Quiet,
```
(both stored as `Quiet`, the available canonical name, not `LowPower`).

`cargo check -p asusd` and `cargo build --release -p asusd` both clean on `ogc/main` tip.

## What we did NOT validate locally

This hardware exposes `quiet balanced performance` only — never `low-power`. So the reverse direction of the bidirectional fallback (a user passing `Quiet` on a hypothetical `low-power`-only machine) was reasoned-through (matches the existing one-direction precedent) but not hardware-tested. The patch is a strict no-op when the requested profile IS in choices, so on dual-choice or LowPower-only hardware the behavior should be unchanged from before this PR.

## Reviewer notes

- Pre-existing behavior left untouched: `set_platform_profile` writes config (line 480 in main) before validating against `choices`. The alias substitution doesn't widen that pre-existing window — the substitution only triggers when the validation would otherwise reject. Worth a separate cleanup PR if desired.
- No tests exist under `#[cfg(test)]` for `set_platform_profile` or the alias paths (verified via `rg`). Happy to add unit tests in a follow-up if maintainers prefer.

### DIFF
--- a/asusd/src/ctrl_platform.rs
+++ b/asusd/src/ctrl_platform.rs
@@ -280,25 +280,32 @@ impl CtrlPlatform {
             self.config.lock().await.platform_profile_on_battery
         };
 
-        // Older configs may still contain Quiet on devices that only support LowPower.
+        // Configs may contain a profile name (Quiet or LowPower) that this
+        // platform does not expose via the kernel ACPI sysfs — different ASUS
+        // platforms expose one or the other for the same semantic profile.
         // Normalize at apply-time so AC/BAT transitions still work correctly.
-        if configured == PlatformProfile::Quiet {
+        let alias = match configured {
+            PlatformProfile::Quiet => Some(PlatformProfile::LowPower),
+            PlatformProfile::LowPower => Some(PlatformProfile::Quiet),
+            _ => None,
+        };
+        if let Some(target) = alias {
             if let Ok(choices) = self.platform.get_platform_profile_choices() {
-                if !choices.contains(&PlatformProfile::Quiet)
-                    && choices.contains(&PlatformProfile::LowPower)
-                {
+                if !choices.contains(&configured) && choices.contains(&target) {
                     let mut cfg = self.config.lock().await;
                     if power_plugged {
-                        cfg.platform_profile_on_ac = PlatformProfile::LowPower;
+                        cfg.platform_profile_on_ac = target;
                     } else {
-                        cfg.platform_profile_on_battery = PlatformProfile::LowPower;
+                        cfg.platform_profile_on_battery = target;
                     }
                     cfg.write();
                     warn!(
-                        "Configured profile Quiet is unavailable, falling back to LowPower for {}",
+                        "Configured profile {:?} is unavailable, falling back to {:?} for {}",
+                        configured,
+                        target,
                         if power_plugged { "AC" } else { "battery" }
                     );
-                    return PlatformProfile::LowPower;
+                    return target;
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


NeroReflex:
Thanks you. This looks like correct but I wonder why have you noticed it? Took the nvme of a laptop and reused in another laptop?

Anyway could you reorganize that code into function(s) and add a few test? Those are becoming too long for my taste.

Ghoul4500 (approved):

PR #135, opened 2026-06-24
Fix: Avoid unnecessary PCI bus scans and reduce GPU detection log noise

idobutidont:
Come to think of it, is the 3s periodic re-detection even really needed?

### DIFF
--- a/asusd/src/ctrl_gpu.rs
+++ b/asusd/src/ctrl_gpu.rs
@@ -168,24 +168,23 @@ impl CtrlGpu {
     /// is available (e.g. dGPU disabled via ASUS attribute), we fall back to
     /// periodic re-detection every 3 seconds to handle hotplug scenarios.
     pub async fn start_watcher(&self, signal_ctxt: SignalEmitter<'static>) -> Result<(), RogError> {
-        let ctrl_inotify = self.clone();
-        let ctrl_fallback = self.clone();
-        let path = self.dgpu_runtime_status_path.clone();
+        let ctrl = self.clone();
 
         tokio::spawn(async move {
-            if let Some(runtime_path) = path {
-                // inotify-based monitoring of the runtime_status sysfs file
-                info!("CtrlGpu: starting inotify watcher on {:?}", runtime_path);
-                let mut ctrl = ctrl_inotify;
-                let mut buffer = [0u8; 32];
+            let mut ctrl = ctrl;
 
-                loop {
+            loop {
+                if let Some(runtime_path) = ctrl.dgpu_runtime_status_path.clone() {
+                    // inotify-based monitoring of the runtime_status sysfs file
+                    info!("CtrlGpu: starting inotify watcher on {:?}", runtime_path);
+                    let mut buffer = [0u8; 32];
+
                     let inotify = match inotify::Inotify::init() {
                         Ok(i) => i,
                         Err(e) => {
                             error!("CtrlGpu: failed to init inotify: {e}");
-                            // Fall back to polling
-                            break;
+                            tokio::time::sleep(std::time::Duration::from_secs(3)).await;
+                            continue;
                         }
                     };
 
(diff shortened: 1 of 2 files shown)


idobutidont:
Well, this is more annoying than I thought...
```
...
asusd[3318]: [INFO  rog_platform::gpu_pci] Found dgpu 10DE:2DD8 at "0000:01:00.0"
asusd[3318]: [INFO  rog_platform::gpu_pci] Found additional device 10DE:22EC at "0000:01:00.1"
...
```

asusd/src/ctrl_gpu.rs
```diff
@@ -164,27 +166,26 @@ impl CtrlGpu {
...
-        let path = self.dgpu_runtime_status_path.clone();
+        let ctrl = self.clone();
 
         tokio::spawn(async move {
-            if let Some(runtime_path) = path {
-                // inotify-based monitoring of the runtime_status sysfs file
-                info!("CtrlGpu: starting inotify watcher on {:?}", runtime_path);
-                let mut ctrl = ctrl_inotify;
-                let mut buffer = [0u8; 32];
+            let mut ctrl = ctrl;
```
Ghoul4500, line 175: This is a bit weird. Why not declare it as mut to begin with?

asusd/src/ctrl_gpu.rs
```diff
@@ -253,20 +249,87 @@ impl CtrlGpu {
...
+                            Err(e) => {
+                                error!("CtrlGpu: failed to match subsystem: {e}");
+                                return false;
+                            }
+                        },
+                        Err(e) => {
+                            error!("CtrlGpu: failed to create MonitorBuilder: {e}");
+                            return false;
+                        }
+                    };
```
Ghoul4500, line 283: This sits inside the outer loop after the inotify branch returns via continue. So this always creates a new monitor and drops it after one event. Maybe create the monitor outside the loop and reuse it across iters.

asusd/src/ctrl_gpu.rs
```diff
@@ -253,20 +249,87 @@ impl CtrlGpu {
...
+                    }
+
+                    loop {
+                        if let Err(e) = poll.poll(&mut events, None) {
+                            error!("CtrlGpu: mio poll failed: {e}");
+                            return false;
+                        }
+                        for event in monitor.iter() {
+                            if let Some(action) = event.action() {
+                                if action.to_str() == Some("add") {
```
Ghoul4500, line 312: Is add really the only event we need to listen for?

asusd/src/ctrl_gpu.rs
```diff
@@ -253,20 +249,87 @@ impl CtrlGpu {
...
+                        &mut monitor,
+                        UDEV,
+                        Interest::READABLE,
+                    ) {
+                        error!("CtrlGpu: failed to register udev monitor with mio: {e}");
+                        return false;
+                    }
+
+                    loop {
+                        if let Err(e) = poll.poll(&mut events, None) {
```
Ghoul4500, line 306: hmmm, this probably needs at least a timeout. It's inside the spawn_blocking. Might be better to use AsyncFd-based wait with tokio net feature

Ghoul4500 (changes requested):
Overall this looks good, thank you for the contribution. It also fixes a dgpu_disabled detection issue which I noticed only a while ago.

Please fix the issues mentioned in the comments (please keep comment history proper, amend commits where needed) and this will be merged

asusd/src/ctrl_gpu.rs
luytan: maybe listen to bind event instead of add ?

Ghoul4500:
Thank you for the contribution and thanks to this we have found several other issues with the surrounding code. I have carried your commits forward and had to drop and change some, and kept your credit where the code remained unchanged. 

Superseded by #214

PR #137, opened 2026-06-25
Fix: Drop unused versions and redundant futures-lite dependencies

scardracs:


NeroReflex:
Always appreciate dropping dependencies. Thanks!

PR #138, opened 2026-06-27
Cleanup/remove legacy crates

scardracs:
Let's continue the good work cleaning up some more legacy crates that are unused/easily replaceable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -229,7 +229,6 @@ dependencies = [
  "argh",
  "dmi_id",
  "env_logger",
- "gumdrop",
  "log",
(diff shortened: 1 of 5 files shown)


Cargo.lock
Ghoul4500, line 1: Remove lockfile pls. Likely to conflict with other PRs so we generate the lockfile once before every release

Ghoul4500 (changes requested):

Cargo.lock
scardracs: Done

Ghoul4500 (approved):

PR #141, opened 2026-06-28
Feat: Add autostart support on settings

scardracs:
## Description

This PR implements the **Autostart** feature for `rog-control-center`. It allows users to configure the application to run automatically on system boot, with the option to launch it minimized to the system tray/background.

### Core Features:
* **System Autostart Integration**: Creates and manages the user's autostart desktop entry file (`~/.config/autostart/rog-control-center.desktop`) automatically.
* **Autostart Customization (UI Closed / Background Startup)**: Provides settings to choose whether the application starts minimized to the background on system startup.
* **Autostart Status Sync**: Dynamically synchronizes the autostart state on application startup to ensure configuration consistency.
* **UI Controls**: Integrates user toggles inside the **App Settings** page to easily enable/disable system autostart and configure its startup state.

---

## Technical Details

- **rog-control-center/src/config.rs**:
  - Added `enable_autostart` and `autostart_in_background` to the core configuration struct.
  - Implemented `update_autostart` to write/remove the `rog-control-center.desktop` entry in `~/.config/autostart/` with correct launch arguments (`--background`).
  - Added unit test coverage for the autostart desktop file creation.
- **rog-control-center/src/cli_options.rs**: Added the `--background` argument flag to support starting minimized.
- **rog-control-center/src/main.rs**: Added startup initialization to keep the autostart entry in sync and parsed the `--background` CLI flag to run the application in the background when autostarted.
- **rog-control-center/ui/pages/app_settings.slint**: Added the `"Start app on system startup"` toggle and the nested `"Start in background on startup"` sub-option.
- **rog-control-center/src/ui/mod.rs**: Registered state bindings and callbacks to handle autostart events from the GUI.

---

## End Result
image

### DIFF
--- a/rog-control-center/src/config.rs
+++ b/rog-control-center/src/config.rs
@@ -13,6 +13,8 @@ pub struct Config {
     pub run_in_background: bool,
     pub startup_in_background: bool,
     pub enable_tray_icon: bool,
+    #[serde(default)]
+    pub enable_autostart: bool,
     pub ac_command: String,
     pub bat_command: String,
     pub dark_mode: bool,
@@ -30,6 +32,7 @@ impl Default for Config {
             run_in_background: true,
             startup_in_background: false,
             enable_tray_icon: true,
+            enable_autostart: false,
             dark_mode: true,
             start_fullscreen: false,
             fullscreen_width: 1920,
@@ -86,6 +89,7 @@ impl From<Config461> for Config {
             run_in_background: c.run_in_background,
             startup_in_background: c.startup_in_background,
             enable_tray_icon: true,
+            enable_autostart: false,
             ac_command: c.ac_command,
             bat_command: c.bat_command,
             dark_mode: true,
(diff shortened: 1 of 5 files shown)


rog-control-center/src/main.rs
```diff
@@ -161,12 +161,14 @@ async fn main() -> Result<()> {
         config.run_in_background = false;
         config.startup_in_background = false;
         config.start_fullscreen = true;
+        config.enable_autostart = false;
     }
 
     config.write();
+    rog_control_center::config::update_autostart(config.enable_autostart, config.autostart_in_background);
```
Ghoul4500: I don't think i'm understanding this. This is run unconditionally on every startup? why

rog-control-center/src/main.rs
```diff
@@ -161,12 +161,14 @@ async fn main() -> Result<()> {
...
 
     let enable_tray_icon = config.enable_tray_icon;
-    let startup_in_background = config.startup_in_background;
+    let startup_in_background = cli_parsed.background || config.startup_in_background;
```
Ghoul4500: What if a user has autostart not in background but has startup in background true? Then autostart will be in background but he never set autostart_in_background to true.  i think that resulting behavior would be confusing. Might be better to not split start_in_background and autostart_in_background. Having them together is cleaner

rog-control-center/src/config.rs
```diff
@@ -96,3 +104,96 @@ impl From<Config461> for Config {
...
+            // Test disabling
+            update_autostart(false, false);
+            assert!(!path.exists());
+
+            // Restore backup if any
+            if let Some(ref backup_content) = backup {
+                std::fs::write(&path, backup_content).ok();
+            }
+        }
+    }
```
Ghoul4500, line 206: If for any reason only half of this test ever runs (error/cancel etc), this would potentially modify the autostart config of anyone running these tests locally. Maybe use a temp dir. You can have the function accept a dir param

Ghoul4500 (changes requested):

rog-control-center/src/main.rs
scardracs: I know it's more convenient to keep them together but, let's say someone wants to enable background startup but not 'open in background' (e.g., you completely closed the app, even from the background, and want to reopen it), it would be inconvenient. That's why I thought it best to keep them separate.

rog-control-center/src/main.rs
scardracs: This runs on startup if you choose the start on startup from the app settings. Have you thought of another way?

scardracs: After reviewing the code again I saw your point. I didn't consider it at all

rog-control-center/src/config.rs
scardracs: Thank for pointing out. Fixed

rog-control-center/src/main.rs
Ghoul4500: Separation is definitely ideal. But the bug I mentioned needs fixing. The easiest solution I could think of was to combine them

scardracs: Ok I thought of a different way. We use only startup_in_background but there are 2 different settings voices that change the cli flags set inside .desktop

Ghoul4500 (approved):
This looks good to me now. Thank you for your contribution

PR #143, opened 2026-06-29
fix(asus-shutdown): exit cleanly on SIGTERM

### LINKED ISSUE #140: Yay wants to restart asusd, but that depends on asus-shutdown, which in turn ignores SIGTERM
I'm running Cachy, and when I update the system it sometimes asks me to restart services, one of which is yours. The problem is it hangs forever, because it waits for asus-shutdown to restart, and it doesn't because you explicitly use `SendSIGKILL=no` on top of a 45s delay, which effectively turns it into a forever delay.

The only things that successfully kills it is a real system shutdown because it listens to a signal from `logind`.

You seem to want to do this to prevent dying in the middle of a firmware write, but it doesn't feel like it's guaranteed because eventually the watchdog will kill you regardless of what systemd wants to do, and logind only really respects the blocking state for 5s, then keeps you running but moves on with the shutdown sequence.

I don't really know what the real solution would be, but right now I did:

```
[Service]
TimeoutStopSec=75
SendSIGKILL=yes
```

So at least it eventually dies.

Maybe a solution could be to actually respect SIGTERM when nothing is in the queue? I don't really know though, I'm not an expert

scardracs:
When asus-shutdown receives a SIGTERM outside of system shutdown (e.g. during manual service management or package update restarts), it ignored the signal and looped indefinitely waiting for logind's PrepareForShutdown signal. Combined with SendSIGKILL=no in the systemd service file, this caused commands like systemctl stop/restart asus-shutdown to hang indefinitely.

This patch makes asus-shutdown break the select loop and exit cleanly upon receiving SIGTERM. Since settings application is run synchronously within the event branch, SIGTERM will not interrupt an ongoing GPU firmware write.

Closes: https://github.com/OpenGamingCollective/asusctl/issues/140
Signed-off-by: Marco Scardovi <scardracs@disroot.org>

### DIFF
--- a/asus-shutdown/src/main.rs
+++ b/asus-shutdown/src/main.rs
@@ -96,8 +96,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     loop {
         tokio::select! {
             _ = sigterm.recv() => {
-                warn!("Received SIGTERM");
-                warn!("Deferring exit until deferred shutdown apply reaches a safe completion point");
+                warn!("Received SIGTERM, exiting cleanly");
+                break;
             }
             event = shutdown_events.next() => {
                 match event {


Ghoul4500 (approved):
Looks correct to me. Thanks

PR #146, opened 2026-07-01
fix: sync Intel WMI CPU power limits to RAPL powercap interfaces dynamically

scardracs:
Sync Intel CPU power limits (`ppt_pl1_spl`, `ppt_pl2_sppt`, `ppt_pl3_fppt`) to the kernel's RAPL (`powercap`) sysfs interfaces.
On modern Intel-based ROG laptops (such as the i9-13980HX on Strix G814JZ), the ASUS BIOS/WMI firmware does not correctly propagate or synchronize WMI power limits to the actual CPU MSR or MMIO RAPL registers. Writing to the WMI attributes under `asus-armoury` succeeds and reads back correctly, but has no actual hardware effect because the kernel's `intel-rapl-mmio` controller remains hard-locked at default constraint wattages (e.g., 45W/90W).
This patch bridges the gap by writing directly to the corresponding `/sys/class/powercap` nodes when Intel limits are set.
## Changes
1. **Intel CPU Detection**: Adds a safe `/proc/cpuinfo` check (`GenuineIntel`) to ensure this logic is only applied on Intel platforms.
2. **RAPL / Powercap Sync**: Walks `/sys/class/powercap` to find any directories belonging to package zones (e.g. `intel-rapl:*` / `intel-rapl-mmio:*`) and dynamically resolves the correct constraint index suffix by scanning and matching `constraint_<idx>_name` against the target name:
   - `"long_term"` for PL1 (`ppt_pl1_spl`)
   - `"short_term"` for PL2 (`ppt_pl2_sppt`)
   - `"peak_power"` for PL3 (`ppt_pl3_fppt`)
3. **Inline Syncing**: Rather than performing a full re-read and sync of all three limits every time, writes are executed dynamically and inline to the single updated attribute's corresponding RAPL constraint when it is modified.
## How to Test
Apply any CPU power limits via `asusctl`:
```bash
asusctl armoury set ppt_pl1_spl 110
asusctl armoury set ppt_pl2_sppt 135
```

### DIFF
--- a/rog-platform/src/asus_armoury.rs
+++ b/rog-platform/src/asus_armoury.rs
@@ -240,9 +240,14 @@ impl Attribute {
                 if let Ok(int) = val.parse::<i32>() {
                     AttrValue::Integer(int)
                 } else if val.contains(';') {
-                    AttrValue::EnumInt(val.split(';').filter_map(|s| s.parse().ok()).collect())
+                    let values: Vec<_> = val.split(';').collect();
+                    if values.iter().all(|v| v.parse::<i32>().is_ok()) {
+                        AttrValue::EnumInt(values.iter().map(|v| v.parse().unwrap()).collect())
+                    } else {
+                        AttrValue::EnumStr(values.iter().map(|s| s.to_string()).collect())
+                    }
                 } else {
-                    AttrValue::EnumStr(val.split(';').map(|s| s.to_string()).collect())
+                    AttrValue::EnumStr(vec![val])
                 }
             }
             Err(_) => AttrValue::None,
--- a/rog-platform/src/asus_armoury.rs
+++ b/rog-platform/src/asus_armoury.rs
@@ -53,6 +53,70 @@ pub struct Attribute {
     base_path: PathBuf,
 }
 
+fn is_intel_cpu() -> bool {
+    std::fs::read_to_string("/proc/cpuinfo")
+        .map(|content| content.contains("GenuineIntel"))
+        .unwrap_or(false)
+}
+
+fn write_intel_rapl_limit(constraint_name: &str, watts: i32) {
+    if !is_intel_cpu() || watts < 0 {
+        return;
+    }
+    let microwatts_str = ((watts as u64) * 1_000_000).to_string();
+    let Ok(entries) = read_dir("/sys/class/powercap") else {
+        return;
+    };
+    for entry in entries.flatten() {
+        let zone_path = entry.path();
+        let zone = entry.file_name();
+        let zone = zone.to_string_lossy();
+        // A package zone name has exactly one ':' ("intel-rapl:0"). Sub-zones have two ("intel-rapl:0:0").
+        let is_package_zone = (zone.starts_with("intel-rapl:")
+            || zone.starts_with("intel-rapl-mmio:"))
+            && zone.matches(':').count() == 1;
+        if !is_package_zone {
+            continue;
+        }
+        let Ok(sub_entries) = read_dir(&zone_path) else {
+            continue;
+        };
+        for sub_entry in sub_entries.flatten() {
+            let file_name = sub_entry.file_name();
+            let file_name = file_name.to_string_lossy();
+            if file_name.starts_with("constraint_") && file_name.ends_with("_name") {
+                let name_path = sub_entry.path();
+                if let Ok(name) = std::fs::read_to_string(&name_path) {
+                    if name.trim() == constraint_name {
+                        let idx_str =
+                            &file_name["constraint_".len()..file_name.len() - "_name".len()];
+                        let limit_path =
+                            zone_path.join(format!("constraint_{}_power_limit_uw", idx_str));
+                        if limit_path.exists() {
+                            debug!("Writing {microwatts_str} to {}", limit_path.display());
+                            match OpenOptions::new().write(true).open(&limit_path) {
+                                Ok(mut file) => {
+                                    if let Err(e) = file.write_all(microwatts_str.as_bytes()) {
+                                        error!(
+                                            "Failed to write to {}: {}",
+                                            limit_path.display(),
+                                            e
+                                        );
+                                    }
+                                }
+                                Err(e) => {
+                                    error!("Failed to open {}: {}", limit_path.display(), e);
+                                }
+                            }
+                        }
+                        break;
+                    }
+                }
+            }
+        }
+    }
+}
+
 impl Attribute {
     pub fn name(&self) -> &str {
         &self.name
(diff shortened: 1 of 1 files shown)


scardracs:
@denisu14 as I don't have an Intel CPU I can't test it. It *should* work but please test it

denisu14:
@scardracs sorry for the late reply. the patch doesn't work for me. i'll investigate and see what fails and let you know.

denisu14:
okay. found something in journald. I had deleted my asusd.ron config and default is false for the tuning field so it disabled this code path.
`asusd[38362]: [WARN  asusd::asus_armoury] Tuning is disabled: skipping setting value to PPT property ppt_pl1_spl`

I thought changing it back to true would fix everything but turns out the patch actually times out the systemd service by infinitely looping and setting pl1 when service starts. 

Fix: https://github.com/scardracs/asusctl/pull/1
Tested and works on my machine.

denisu14:
We may want to add a way to set `ac_profile_tunings` via `asusctl` as some people may not be aware of the config option in `/etc/asusd.ron`.

scardracs:
> We may want to add a way to set `ac_profile_tunings` via `asusctl` as some people may not be aware of the config option in `/etc/asusd.ron`.

I think https://github.com/OpenGamingCollective/asusctl/pull/142 will make everything easier once merged

scardracs:
> okay. found something in journald. I had deleted my asusd.ron config and default is false for the tuning field so it disabled this code path. `asusd[38362]: [WARN asusd::asus_armoury] Tuning is disabled: skipping setting value to PPT property ppt_pl1_spl`
> 
> I thought changing it back to true would fix everything but turns out the patch actually times out the systemd service by infinitely looping and setting pl1 when service starts.
> 
> Fix: [scardracs#1](https://github.com/scardracs/asusctl/pull/1) Tested and works on my machine.

Yeah, way better than my implementation. Merged on my branch

denisu14:
seems like force push deleted the fix commit. 28dffa0

scardracs:
> seems like force push deleted the fix commit. [28dffa0](https://github.com/OpenGamingCollective/asusctl/commit/28dffa0b9fb0f77673716ed96fca461c0a9d9565)

I see I forgot to rebase my local with it. Done

Ghoul4500:
Why re-read and sync all 3 limits separately? We already have the value in hand when writing to the ppt_* attribute in the first place, so you could write the corresponding powercap file at the same point.

Also one thing to note, the indexes 0,1,2 for spl 1 2 3 is not guaranteed. Use `/sys/class/powercap/<zone>/constraint_<N>_name` instead. It's a more reliable way to see long_term for pl1, short_term for pl2 and peak_power for pl3 instead of relying on an index that may change per machine.

scardracs:
> Why re-read and sync all 3 limits separately? We already have the value in hand when writing to the ppt_* attribute in the first place, so you could write the corresponding powercap file at the same point.
> 
> Also one thing to note, the indexes 0,1,2 for spl 1 2 3 is not guaranteed. Use `/sys/class/powercap/<zone>/constraint_<N>_name` instead. It's a more reliable way to see long_term for pl1, short_term for pl2 and peak_power for pl3 instead of relying on an index that may change per machine.

You are right. To make sure it won't cause problems in the future I've made the program scan to actually check how many pl are present dinamically and also made that the ppts and pls correspond.

scardracs:
As I was at it I noticed a possible bug at possible_values so I planned to fix it as well in the meantime.
The parsing of possible_values was incorrect.

The current code were as follow:
```bash
if let Ok(int) = val.parse::<i32>() {
    AttrValue::Integer(int)
} else if val.contains(';') {
    AttrValue::EnumInt(
        val.split(';').filter_map(|s| s.parse().ok()).collect()
    )
} else {
    AttrValue::EnumStr(...)
}
```
Now imagine:
```bash
possible_values -> low;medium;high
```

The code enters
```bash
val.contains(';')
```

and produces
```bash
EnumInt(vec![])
```

because every parse fails the string enum is lost.

Ghoul4500:
> As I was at it I noticed a possible bug at possible_values so I planned to fix it as well in the meantime. The parsing of possible_values was incorrect.
> 
> The current code were as follow:
> 
> ```shell
> if let Ok(int) = val.parse::<i32>() {
>     AttrValue::Integer(int)
> } else if val.contains(';') {
>     AttrValue::EnumInt(
>         val.split(';').filter_map(|s| s.parse().ok()).collect()
>     )
> } else {
>     AttrValue::EnumStr(...)
> }
> ```
> 
> Now imagine:
> 
> ```shell
> possible_values -> low;medium;high
> ```
> 
> The code enters
> 
> ```shell
> val.contains(';')
> ```
> 
> and produces
> 
> ```shell
> EnumInt(vec![])
> ```
> 
> because every parse fails the string enum is lost.

This belongs in a separate commit then. Thank you for noting it

rog-platform/src/asus_armoury.rs
```diff
@@ -53,6 +53,59 @@ pub struct Attribute {
...
+    std::fs::read_to_string("/proc/cpuinfo")
+        .map(|content| content.contains("GenuineIntel"))
+        .unwrap_or(false)
+}
+
+fn write_intel_rapl_limit(constraint_name: &str, watts: i32) {
+    if !is_intel_cpu() {
+        return;
+    }
+    let microwatts_str = ((watts as u64) * 1_000_000).to_string();
```
Ghoul4500, line 66: Since this is cast from an i64 to u64, add a check to make sure watts >= 0

rog-platform/src/asus_armoury.rs
```diff
@@ -53,6 +53,59 @@ pub struct Attribute {
...
+                if let Ok(name) = std::fs::read_to_string(&name_path) {
+                    if name.trim() == constraint_name {
+                        let idx_str =
+                            &file_name["constraint_".len()..file_name.len() - "_name".len()];
+                        let limit_path =
+                            zone_path.join(format!("constraint_{}_power_limit_uw", idx_str));
+                        if limit_path.exists() {
+                            debug!("Writing {microwatts_str} to {}", limit_path.display());
+                            if let Ok(mut file) = OpenOptions::new().write(true).open(&limit_path) {
+                                let _ = file.write_all(microwatts_str.as_bytes());
```
Ghoul4500: Instead of discarding this, log debug on fail. Otherwise a failure here would still be silent

Ghoul4500 (changes requested):
I still want some input from @NeroReflex for this one. I'm unsure if we should even be writing this firmware attribute in the first place. It would fight other tools like thermald as well and the change will be silent as we are not actively monitoring that file

Ghoul4500:
> I still want some input from @NeroReflex for this one. I'm unsure if we should even be writing this firmware attribute in the first place. It would fight other tools like thermald as well and the change will be silent as we are not actively monitoring that file

It feels to me that we are dealing with a kernel bug and trying to put a bandage over it in userspace instead of fixing the root cause. I'll have to look into this bug a bit more

scardracs:
I feel the same way. Maybe if Ilpo and other intel guys can give their opinion it would be helpful

scardracs:
Gonna close this one as it needs to be done kernel/bios side

PR #164, opened 2026-07-14
Update About page

Piedfer:
- Change source code link to the OGC Github
- Add @tr macros for translation
- Update mains .po and .mo files

Do you want to specify in the requirements that it is preferable to have the latest kernel? If yes, how?


Ghoul4500:
> * Change source code link to the OGC Github
> * Add @tr macros for translation
> * Update mains .po and .mo files
> 
> Do you want to specify in the requirements that it is preferable to have the latest kernel? If yes, how?

The kernel requirement is 6.19 or later. The latest kernel is always the recommended kernel

Ghoul4500 (changes requested):
Other than the suggested change, if possible, I would appreciate if you could verify the effect this has on other translations (since there's some reordering and stuff). If possible, update whatever you can confidently do in separate commits

rog-control-center/ui/pages/about.slint
```diff
@@ -42,7 +42,7 @@ export component PageAbout inherits VerticalLayout {
                 }
 
                 Text {
-                    text: @tr("      • This software requires kernel version 6.19.");
+                    text: @tr("      • The kernel requirement is 6.19 or later. The latest kernel is always the recommended kernel.");
```
Ghoul4500: this could be worded better imo. maybe something like

"This software requires kernel version 6.19 or later. However due to active development, the latest kernel is always the recommended one."

Ghoul4500 (changes requested):

rog-control-center/ui/pages/about.slint
Piedfer: Understood. I get back to it when I have a little more time. For everything.
For now old and new Frensh translation files work well with a new pot.

Ghoul4500:
@Piedfer please fix th commit history on this

PR #185, opened 2026-07-21
fix(asusd-user): wait dynamically for system daemon on D-Bus

scardracs:
## Problem & Symptoms
On several ASUS ROG laptop models (e.g., Strix SCAR 16 G635LW) using version 6.3.9, special hotkeys (Volume Up/Down, Mic Mute, Fan profile speed toggle, ROG Key) ceased working after system boot.

## Root Cause
In version 6.3.9, the system daemon (`asusd`) performs several complex/asynchronous initialization tasks on startup (such as default fan curve ACPI queries, XG Mobile LED checks, and D-Bus firmware attribute registration), which noticeably prolonged its overall boot sequence. The user-level daemon (`asusd-user`) was configured to start blindly after a 2-second sleep (`ExecStartPre=/usr/bin/sleep 2` in the systemd user service). Because `asusd` took longer than 2 seconds to register its name (`xyz.ljones.Asusd`) on D-Bus:
1. `asusd-user` failed to query the available D-Bus interfaces and crashed on start.
2. Under systemd's default restart policies (`StartLimitBurst=2` in 200 seconds), `asusd-user` exhausted its startup attempts after two failures and was permanently disabled for the rest of the boot session, breaking all special hotkeys.
Furthermore, keeping `StartLimitBurst=2` meant that if the system daemon was restarted manually a couple of times during a session, the user daemon would crash (due to D-Bus disconnection) and quickly hit the restart limits, leaving the service disabled permanently until manually restarted.

## Proposed Changes
This PR resolves the startup race condition and improves the overall robustness of the user service:
1. **Dynamic D-Bus Wait in `asusd-user`** (Commit 1):
   - Replaced the immediate startup crash in `list_iface_blocking()` at boot with a deterministic check of the D-Bus name owner status.
   - If `xyz.ljones.Asusd` has no owner yet, the daemon subscribes to the D-Bus system daemon's `NameOwnerChanged` signal and blocks dynamically.
   - Once `asusd` registers its name and becomes ready, `asusd-user` breaks the loop and continues its initialization. This is fully event-driven and resource-friendly.

2. **Systemd Service Clean Up & Restart Limits Removal** (Commit 2):
   - Removed the obsolete `ExecStartPre=/usr/bin/sleep 2` from `data/asusd-user.service`, enabling the service to spawn instantly and wait cleanly in code.
   - Removed the aggressive `StartLimitInterval=200` and `StartLimitBurst=2` options, delegating rate-limiting of restarts to standard systemd defaults (e.g., 5 restarts in 10 seconds), preventing the daemon from getting stuck in an inactive state.

## Commits
* `fix(asusd-user): wait dynamically for system daemon on D-Bus` — Updates `asusd-user/src/daemon.rs` with the dynamic wait logic.
* `fix(systemd): clean up user service startup and restart limits` — Updates `data/asusd-user.service` to remove the sleep workaround and the start limit constraints.


asusd-user/src/daemon.rs
```diff
@@ -36,6 +36,20 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let name = zbus::names::BusName::try_from(DBUS_NAME)?;
+    if !dbus.name_has_owner(name.clone())? {
+        println!("  waiting for system daemon to become ready on D-Bus...");
+        let mut stream = dbus.receive_name_owner_changed()?;
+        for signal in stream.by_ref() {
+            let args = signal.args()?;
+            if args.name() == &name && args.new_owner().is_some() {
+                println!("  system daemon is ready!");
+                break;
+            }
```
Ghoul4500: this should be info macro and not println

Ghoul4500 (changes requested):
Is there any reason why we can't simply do something like

```
[Unit]
After=asusd.service
Wants=asusd.service
```

if we are doing any waiting though, adding timeout is also important

scardracs:
> Is there any reason why we can't simply do something like
> 
> ```
> [Unit]
> After=asusd.service
> Wants=asusd.service
> ```
> 
> if we are doing any waiting though, adding timeout is also important

A user-level systemd instance cannot see or wait for system-level services directly in `After=`. A user service can only see other user services.

asusd-user/src/daemon.rs
```diff
@@ -29,13 +30,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
+    let dbus = zbus::blocking::fdo::DBusProxy::new(&conn)?;
+    let name = zbus::names::BusName::try_from(DBUS_NAME)?;
+    if !dbus.name_has_owner(name.clone())? {
+        info!("  waiting for system daemon to become ready on D-Bus...");
+        let (tx, rx) = std::sync::mpsc::channel();
+        let dbus_clone = dbus.clone();
+        let name_clone = name.clone();
+        std::thread::spawn(move || {
+            if let Ok(mut stream) = dbus_clone.receive_name_owner_changed() {
```
Ghoul4500: I think there will need to be some improvements to this part, but I'll need to think a bit about what really needs to be done here

asusd-user/src/daemon.rs
```diff
@@ -29,13 +30,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                            let _ = tx.send(());
+                            break;
+                        }
+                    }
+                }
+            }
+        });
+
+        if rx.recv_timeout(std::time::Duration::from_secs(30)).is_err() {
+            warn!("Timed out waiting for system daemon to become ready on D-Bus");
```
Ghoul4500: this should be error. not warn

Ghoul4500 (changes requested):

asusd-user/src/daemon.rs
```diff
@@ -29,13 +30,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                for signal in stream.by_ref() {
+                    if let Ok(args) = signal.args() {
+                        if args.name() == &name_clone && args.new_owner().is_some() {
+                            let _ = tx.send(());
+                            break;
+                        }
+                    }
+                }
+            }
+        });
```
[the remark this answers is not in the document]
Ghoul4500: Yea that seems about right

asusd-user/src/daemon.rs
```diff
@@ -29,13 +30,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    println!("  user daemon v{}", asusd_user::VERSION);
-    println!("    rog-anime v{}", rog_anime::VERSION);
-    println!("     rog-dbus v{}", rog_dbus::VERSION);
-    println!("rog-platform v{}", rog_platform::VERSION);
+    info!("  user daemon v{}", asusd_user::VERSION);
+    info!("    rog-anime v{}", rog_anime::VERSION);
+    info!("     rog-dbus v{}", rog_dbus::VERSION);
+    info!("rog-platform v{}", rog_platform::VERSION);
 
     let conn = zbus::blocking::Connection::system().unwrap();
```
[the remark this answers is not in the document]
scardracs: Better use system()? Instead of system().unwrap() ?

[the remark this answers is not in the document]
Ghoul4500: I'd say it's better in this case. Otherwise handle the error there urself

asusd-user/src/daemon.rs
```diff
@@ -88,16 +146,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        let aura_proxy_blocking = AuraProxyBlocking::new(&conn).unwrap();
+        let aura_proxy_blocking = AuraProxyBlocking::new(&conn)?;
         tokio::task::spawn_blocking(move || loop {
             aura_config.aura.next_state(&layout);
             let packets = aura_config.aura.create_packets();
 
-            aura_proxy_blocking.direct_addressing_raw(packets).unwrap();
+            if let Err(e) = aura_proxy_blocking.direct_addressing_raw(packets) {
+                error!("Aura direct addressing error: {e}");
+            }
```
Ghoul4500, line 161: add a break here lol. how'd i miss this before

Ghoul4500 (changes requested):
this last change fr

Ghoul4500 (approved):

PR #188, opened 2026-07-21
feat: add FA608WI aura support

RPochyly:
aura_support.ron entry for FA608PM is the same for FA608WI.


Ghoul4500 (approved):

PR #191, opened 2026-07-22
fix: allow module loading and remove privileged system call restrictions in asus-shutdown.service

### LINKED ISSUE #184: Brightness control breaks after switching GPU modes

scardracs:
## Summary

Fixes GPU mode switching failures (e.g. switching between Hybrid and Integrated modes via `asusctl armoury set dgpu_disable 1`) by resolving seccomp sandbox restrictions and shutdown inhibitor deadlocks in `asus-shutdown`.

## Root Cause Analysis

1. **Systemd Seccomp Restriction (`SIGSYS`)**:
   `asus-shutdown.service` had restrictive `SystemCallFilter` rules and lacked capabilities (`CAP_SYS_ADMIN`, `CAP_SYS_MODULE`), causing `modprobe -r` to be killed with `SIGSYS (core dumped)` when attempting to unload kernel modules.

2. **Shutdown Inhibitor Deadlock (`Module nvidia is in use`)**:
   `asus-shutdown` holds systemd's *shutdown delay inhibitor* while applying deferred GPU attributes. Previously, `asus-shutdown` waited passively for `nvidia-powerd.service` to stop. However, systemd delays unit teardown while inhibitors are active. This created a deadlock where `nvidia-powerd` remained active, holding `/dev/nvidia*` file descriptors open and causing `modprobe -r` to fail with `FATAL: Module nvidia is in use`, which in turn prevented ACPI firmware attribute writes (`dgpu_disable`).

## Changes Made

This PR splits the fix into two distinct commits:

### 1. `fix(systemd): allow module unloading and system calls in asus-shutdown.service`
- Updated `CapabilityBoundingSet` to include `CAP_SYS_MODULE` and `CAP_SYS_ADMIN`.
- Updated `SystemCallFilter` to allow `@module` and removed restrictions blocking kernel module management.

### 2. `fix(asus-shutdown): stop NVIDIA services and retry module unload during shutdown`
- **Explicit Service Teardown**: `asus-shutdown` now explicitly issues `systemctl stop` for NVIDIA services (`nvidia-powerd.service`, `nvidia-persistenced.service`, `nvidia-fabricmanager.service`) prior to module unloading.
- **Unload Retry Loop**: Added a 5-attempt retry loop (with a 500ms delay) for `modprobe -r` to allow driver file descriptors to settle cleanly.
- **Expanded GPU Node Monitoring**: Extended `busy_gpu_nodes()` detection to track open file handles on `/dev/nvidia*` in addition to `/dev/dri/`.
- **Cleanup**: Removed unused constant definitions and the redundant wait loop for `systemd-logind.service`.

## Testing

Tested on an **ASUS ROG Strix G16 (G614PR - 2025)** running CachyOS (Arch Linux):
- Verified `journalctl -b -1 -u asus-shutdown.service`: all 6 shutdown phases complete in 0 seconds with zero errors.
- Verified NVIDIA modules unload cleanly, ACPI firmware attributes (`dgpu_disable` and `gpu_mux_mode`) apply successfully, and `asusctl armoury list` reflects the updated GPU state on the subsequent boot.

Fixes: https://github.com/OpenGamingCollective/asusctl/issues/184

### DIFF
--- a/asus-shutdown/src/main.rs
+++ b/asus-shutdown/src/main.rs
@@ -22,15 +22,12 @@ const SHUTDOWN_REASON: &str = "defer risky ASUS GPU firmware writes until shutdo
 const WAIT_FOR_GPU_IDLE: Duration = Duration::from_secs(15);
 const GPU_IDLE_POLL_INTERVAL: Duration = Duration::from_millis(500);
 const GPU_IDLE_STABLE_FOR: Duration = Duration::from_secs(2);
-const WAIT_FOR_NVIDIA_POWERD_EXIT: Duration = Duration::from_secs(10);
-const WAIT_FOR_LOGIND_EXIT: Duration = Duration::from_secs(20);
 const ASUSD_BUS_NAME: &str = "xyz.ljones.Asusd";
 const ASUSD_ARMOURY_IFACE: &str = "xyz.ljones.AsusArmoury";
 const SYSTEMD1_BUS_NAME: &str = "org.freedesktop.systemd1";
 const SYSTEMD1_MANAGER_PATH: &str = "/org/freedesktop/systemd1";
 const SYSTEMD1_MANAGER_IFACE: &str = "org.freedesktop.systemd1.Manager";
 const SYSTEMD1_UNIT_IFACE: &str = "org.freedesktop.systemd1.Unit";
-const LOGIND_SERVICE: &str = "systemd-logind.service";
 const NVIDIA_POWERD_SERVICE: &str = "nvidia-powerd.service";
 const NVIDIA_SERVICES: &[&str] = &[
     NVIDIA_POWERD_SERVICE,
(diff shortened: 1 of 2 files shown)


Ghoul4500 (approved):

From my own testing on a G835LW, this fixes the integrated GPU mode and allows me to use it, but detecting that I am currently in integrated mode fails and rogcc claims I am in Hybrid mode. That issue is a pre-existing behaviour so it's just something to consider as a follow-up to this PR as I will be merging this as this brings the functionality

PR #194, opened 2026-07-23
Remove metainfo from asusctl package to resolve conflict with rog-control-center package

Ghoul4500:


ayush123460:
Insane, I forgot we have the `_package` sections :sweat_smile: LGTM!

PR #212, opened 2026-07-27
Fix: Pause Telemetry Loop When Window is Hidden

### LINKED ISSUE #208: ROGCC: dGPU is kept alive
Hi,

the latest version 6.3.10 keeps my dGPU active. I tested it with the following command:

```
watch -n1 "cat /sys/bus/pci/devices/0000:01:00.0/power/runtime_status"
```

It switches to `active` as soon as ROGCC is starting and switches back to `suspended` after ~15 seconds after ROGCC was closed. The problem doesn't exist in 6.3.8.

* OS: CachyOS
* Kernel: 7.1.4-1-cachyos
* Architecture: x86-64

My assumption is that the dGPU is kept alive by the function to show the dGPU temperature.

Issue #180 is not related, because this issue is about the dGPU active status - not the dGPU Hybrid/etc... status.

EDIT: Clarification regarding #180

ayush123460:
The telemetry loop continues to poll in the background even when the window is not visible, leading to unnecessary hold-up of the dGPU leading it to not go into power-saving (d3cold). 

I am taking Arc reference to the AppState and checking if the window is open, and skipping the iteration if the window is not.

There is 1 case where this will fail: if the window is visible but minimized to the "app bar" / taskbar, then polling still continues. Checking if the window is not minimized will require us to hook into Slint's visibility mechanic, and I did not feel that's required here.

Fixes #208

### DIFF
--- a/rog-control-center/src/main.rs
+++ b/rog-control-center/src/main.rs
@@ -236,7 +236,7 @@ async fn main() -> Result<()> {
             if is_rog_ally {
                 let config_copy_2 = config.clone();
                 let newui =
-                    setup_window(config.clone(), prefetched_supported.clone(), is_tuf, None);
+                    setup_window(config.clone(), prefetched_supported.clone(), app_state.clone(), is_tuf, None);
                 newui.window().on_close_requested(move || {
                     exit(0);
                 });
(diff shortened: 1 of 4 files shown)


Ghoul4500 (approved):
Very nicely done!

PR #215, opened 2026-07-27
fix(rog-control-center): move AsusPower initialization outside telemetry loop to fix FD leak

scardracs:


Ghoul4500 (approved):

PR #217, opened 2026-07-27
feat(asusd): add PPD conflict detection warning

### LINKED ISSUE #205: asusd: no warning when power-profiles-daemon is running — PPD contention silently resets EPP to amd_pstate default (was: profile_balanced_epp ignored)
### Description

`profile_balanced_epp` in `asusd.ron` is honoured when asusd applies the Balanced profile **without a platform-profile transition**, but is ignored when asusd **transitions into** Balanced from another profile. In the transition case the EPP written to sysfs is `balance_performance` regardless of the configured value.

Because `change_platform_profile_on_ac`/`_on_battery` are `true` by default, every AC plug/unplug goes through the transition path — so in normal daily use the configured Balanced EPP is effectively never applied.

Notably, asusd logs `Setting BalancePower` (twice) in the failing case, yet the value that lands in sysfs is `balance_performance`. The debug log and the effective write disagree.

### Environment

- **asusctl/asusd:** 6.3.10-1 (daemon self-reports `6.3.9` via the `Version` D-Bus property)
- **Board:** ASUS ROG Strix G16 G614FR (BIOS G614FR.316)
- **CPU:** AMD Ryzen 9 9955HX3D (16C/32T)
- **Kernel:** 7.1.4-1-cachyos
- **Distro:** CachyOS
- **Scaling driver:** `amd-pstate-epp`, `amd_pstate/status` = `active`
- Relevant config:
  ```
  platform_profile_linked_epp: true,
  platform_profile_on_ac: Balanced,
  change_platform_profile_on_ac: true,
  profile_balanced_epp: BalancePower,
  profile_performance_epp: Performance,
  ```

### Steps to reproduce

Deterministic, two commands. Requires `profile_balanced_epp` set to something other than `BalancePerformance` (here `BalancePower`).

```bash
# Baseline: restart asusd while the profile is ALREADY Balanced.
sudo systemctl restart asusd; sleep 6
cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
# -> balance_power        CORRECT, matches profile_balanced_epp

asusctl profile set Performance; sleep 3
cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
# -> performance          correct

asusctl profile set Balanced; sleep 3
cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
# -> balance_performance  WRONG, expected balance_power
```

The same divergence occurs on daemon start depending on whether a transition is required:

| asusd condition | `Available EPP` logged | EPP in sysfs | Correct? |
|---|---|---|---|
| restart, profile already `balanced` | full list | `balance_power` | ✅ |
| restart, transitions `performance` → `balanced` | `[Performance]` | `balance_performance` | ❌ |
| `asusctl profile set Performance` | full list | `performance` | ✅ |
| `asusctl profile set Balanced` | `[Performance]` then full list | `balance_performance` | ❌ |

### Expected

Applying the Balanced profile writes the configured `profile_balanced_epp` (`balance_power`) to `energy_performance_preference`, whether or not a platform-profile transition was required.

### Actual

On a transition into Balanced, `balance_performance` is written instead.

### Logs

Failing case — `asusctl profile set Balanced` (`RUST_LOG=asusd=debug`):

```log
[INFO  asusd::ctrl_platform] ThrottlePolicy setting EPP
[DEBUG asusd::ctrl_platform] Available EPP: [Performance]
[WARN  asusd::ctrl_platform] powersave governor is not is use, trying to set.
[INFO  asusd::ctrl_platform] ThrottlePolicy setting EPP
[DEBUG asusd::ctrl_platform] Available EPP: [Default, Performance, BalancePerformance, BalancePower, Power, Default]
[DEBUG asusd::ctrl_platform] Setting BalancePower
[INFO  asusd::ctrl_platform] ThrottlePolicy setting EPP
[DEBUG asusd::ctrl_platform] Available EPP: [Default, Performance, BalancePerformance, BalancePower, Power, Default]
[DEBUG asusd::ctrl_platform] Setting BalancePower
```
→ resulting sysfs value: `balance_performance`

Working case — restart with profile already `balanced`:

```log
[DEBUG asusd::ctrl_platform] Setting Balanced before EPP
[INFO  asusd::ctrl_platform] ThrottlePolicy setting EPP
[DEBUG asusd::ctrl_platform] Available EPP: [Default, Performance, BalancePerformance, BalancePower, Power, Default]
[DEBUG asusd::ctrl_platform] Setting BalancePower
```
→ resulting sysfs value: `balance_power`

Two log signatures appear **only** in the failing runs and never in the working ones:

1. `Available EPP: [Performance]` — a truncated enumeration containing a single entry.
2. `[WARN] powersave governor is not is use, trying to set.` (typo verbatim from source.)

### Ruled out: the kernel is not the actor

My first hypothesis was that the kernel's `amd-pstate` `platform_profile` handler resets EPP after asusd's write. **Tested and refuted** — with asusd stopped, EPP is unaffected by direct `platform_profile` writes:

```bash
sudo systemctl stop asusd
echo balance_power | sudo tee /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_preference
echo performance   | sudo tee /sys/firmware/acpi/platform_profile   # -> EPP still balance_power
echo balanced      | sudo tee /sys/firmware/acpi/platform_profile   # -> EPP still balance_power
```

So the `balance_performance` value originates from asusd, not from a kernel-side reset.

### Secondary observation: EPP enumeration parse

The enumerated list logs `Default` twice:

```
Available EPP: [Default, Performance, BalancePerformance, BalancePower, Power, Default]
```

against five sysfs values:

```
$ cat /sys/devices/system/cpu/cpu0/cpufreq/energy_performance_available_preferences
default performance balance_performance balance_power power custom
```

`custom` appears to fall back to `Default` rather than being recognised or skipped. Probably harmless on its own, but it means the enumeration is not a faithful representation of what the driver offers — and the truncated `[Performance]` list in the failing path suggests this enumeration is worth a look generally.

### Workaround

Restart asusd while the profile already matches the configured one, so no transition is required:

```bash
sudo systemctl restart asusd   # lands Balanced (a transition, wrong EPP)
sudo systemctl restart asusd   # now no transition needed, correct EPP
```

### Possibly related

#136 (fan curves desync from platform profile after boot) is a different symptom but sits in the same `ctrl_platform` profile-application path, and its log excerpt shows the same `Setting <profile> before EPP` sequencing. Both look like per-profile state not being fully applied on a profile transition; they may share a root cause.

scardracs:
## Summary

This Pull Request resolves power profile contention between `power-profiles-daemon` (PPD) and `asusd` while providing seamless native compatibility for **KDE Plasma** and **GNOME** desktop power widgets.

### Problem
When `power-profiles-daemon` and `asusd` run concurrently, PPD reacts to `platform_profile` changes made by `asusd` and silently resets CPU EPP to driver defaults (`balance_performance`), overriding configured EPP settings (`profile_balanced_epp`).

### Solution
**PPD Conflict Warning**: `asusd`, `rog-control-center`, and `asusctl` check at startup if `net.hadess.PowerProfiles` is already owned on system D-Bus. If active, a prominent warning is logged instructing the user to disable `power-profiles-daemon.service`.

---

## Detailed Changes
* **`rog-control-center/src/main.rs` & `asusctl/src/main.rs`**: Added startup PPD conflict warning checks.
* **`MANUAL.md` & `distro-packaging/asusctl.install`**: Updated manual documentation and installation scripts.

---

## Verification & Testing
- [x] **`cargo check --all-targets`**: Clean compilation across workspace crates.
- [x] **`cargo test --all`**: All unit tests passing (100% pass rate).
- [x] **`cargo clippy --all -- -D warnings`**: 0 warnings.
- [x] **`cargo cranky`**: 0 warnings.
- [x] **`cargo fmt --all -- --check`**: Formatting verified.
- [x] **Git Hooks**: Pre-commit and pre-push hooks executed cleanly.

---

Closes #205

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -71,7 +71,27 @@ These options are not written to the config file as they are stored in efivars. 
 
 ### Profiles
 
-asusctl can support setting a power profile via platform_profile drivers. This requires [power-profiles-daemon](https://gitlab.freedesktop.org/hadess/power-profiles-daemon) v0.10.0 minimum. It also requires the kernel patch for platform_profile support to be applied form [here](https://lkml.org/lkml/2021/8/18/1022) - this patch is merged to 5.15 kernel upstream.
+asusctl supports setting power profiles via ACPI `platform_profile` drivers.
+
+> [!IMPORTANT]
+> **Compatibility with Power Profiles Daemons**:
+> Running an external power profiles daemon (such as `power-profiles-daemon` or `tuned`) concurrently with `asusd`'s profile management can cause race conditions and contention over `/sys/firmware/acpi/platform_profile` and CPU EPP preferences.
+> 
+> You have two options for managing power profiles:
+> 
+> 1. **Allow `asusd` to manage profiles**: Disable the conflicting service (e.g. `power-profiles-daemon`):
+> 
+>    ```bash
+>    sudo systemctl disable --now power-profiles-daemon.service
+>    ```
+> 
+> 2. **Allow an external daemon (PPD/Tuned) to manage profiles**: Keep `power-profiles-daemon` active, and disable `asusd`'s profile management settings in `/etc/asusd/asusd.ron`:
+> 
+>    ```ron
+>    change_platform_profile_on_ac: false,
+>    change_platform_profile_on_battery: false,
+>    platform_profile_linked_epp: false,
+>    ```
 
 A common use of asusctl is to bind the `fn+f5` (fan) key to `asusctl profile -n` to cycle through the 3 profiles:
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,10 @@ Due to ongoing driver work, the minimum suggested kernel version is always **the
 
 Support for TDP is tied to the new asus-armoury driver: available mainline since Linux 6.19: everything older is not supported.
 
+## Power Profiles
+
+asusctl supports and manage power profile daemons. Be sure to follow [Manual](MANUAL.md) on how to manage it and warnings about other ppds.
+
 ## X11 support
 
 X11 is not, and will not, supported by asusctl in any way. We will not help you with X11 issues if there are any due to limited time and it being unmaintained itself. You can however build `rog-control-center` with it enabled `cargo build --features "rog-control-center/x11"`.


qwertyman100:
**Reporter of #205 here.** Thanks for the quick turnaround on this @scardracs, and thanks @Ghoul4500 for weighing in on scope.

One note on testing first: `RUST_LOG=debug cargo run -p rog-control-center` only exercises the new warning block in the GUI — the `net.hadess.PowerProfiles` provider lives in `asusd`, so that run can't reach it. On my machine PPD is currently active and owns the name, so the warning is the only path that would fire. I'm happy to run it, but I want to flag something in the provider first, because I think it bears directly on @Ghoul4500's "we don't want to replace PPD and have to maintain that" concern.

### `HoldProfile` / `ReleaseProfile` are stubs

In `asusd/src/ppd_dbus.rs`, `hold_profile()` returns cookie `1` without applying or tracking anything, `release_profile()` is a no-op, and `ActiveProfileHolds` is hardcoded to `[]`.

Holds aren't a corner of the PPD interface — they're how `powerprofilesctl launch`, Steam/Lutris wrappers, GameMode and GNOME/KDE's own transient boosts all work. Every one of those calls `HoldProfile`, gets a success cookie back, and relies on the daemon releasing the hold when the client's D-Bus connection drops. Against this implementation they would all silently do nothing: no error, no profile change, no visible hold, nothing in any log the user would think to read.

That's the same silent-wrong-outcome failure mode that made the original bug in #205 so hard to diagnose — asusd wrote the right value, lost the race, and the only symptom was a wrong sysfs number.

For context on how visible this would be in practice: my own setup has 12 Steam titles plus a Heroic global wrapper running through `powerprofilesctl launch -p performance -- %command%`. Under this provider all of them would quietly stop boosting, and the only symptom would be worse frame rates.

### Three smaller things

1. **No `PropertiesChanged` for `ActiveProfile` on out-of-band changes.** Nothing emits when the profile changes via `fn+F5`, `asusctl profile set`, or asusd's own `change_platform_profile_on_ac` AC/battery switching. PPD signals on every change, so a desktop widget would show a stale profile — visible on any plug/unplug, which is the most common profile change there is.
2. **EPP is written before the profile is validated** in `set_platform_profile_from_ppd()`: `check_and_set_epp()` and `config.write()` both run *before* the `choices.contains(&policy)` check, so an unsupported profile returns `NotSupported` having already changed EPP and persisted config.
3. **D-Bus name-ownership policy.** The system bus denies name ownership by default; `net.hadess.PowerProfiles` is ownable only because the *power-profiles-daemon package* ships the `system.d` policy file permitting it. If a user follows the new `MANUAL.md` guidance and removes PPD rather than just disabling it, that file goes too and asusd's `request_name` fails — the code `warn!`s and continues, so the feature quietly stops working. This direction would need asusctl to ship its own policy file.

### Suggestion: split this PR

The warning half and the provider half have very different risk profiles, and right now they're bundled so the easy one is blocked by the hard one.

The **warning** is ~35 lines across three binaries, uncontroversial, and closes #205 outright — my residual ask on that issue was only ever a startup warning, and it would have saved me the several hours that produced the two corrections in that thread. That could merge now.

The **`net.hadess.PowerProfiles` provider** is the part that needs the discussion @Ghoul4500 described. Implementing PPD's semantics properly — hold refcounting, per-client tracking, release-on-disconnect, `ActiveProfileHolds` signalling, plus the D-Bus policy and packaging — is a substantially larger and *ongoing* commitment, since it then has to stay in step with PPD's own evolution. That's exactly the maintenance burden being weighed, and I don't think the current diff's scope reflects its true size.

Splitting would let the small win land while the larger question stays open on its own merits.

Happy to test anything useful on a G614FR (Ryzen 9 9955HX3D / RTX 5070 Ti, CachyOS, KDE Plasma) — including the warning path as-is, and the provider if the hold semantics get filled in.

qwertyman100:
Thanks for splitting it @scardracs — this is now exactly what I was asking for in #205, and I'd be happy to see it merged as-is. I'm building the branch now and will report the warning output.

Two things the split left behind, both in the docs rather than the code:

### 1. Unterminated code fence in `MANUAL.md`

The closing fence was removed along with the provider sentence:

````
> ```bash
> sudo systemctl disable --now power-profiles-daemon.service

A common use of asusctl is to bind the `fn+f5` (fan) key...
````

The block never closes.

### 2. The manual still tells users to disable PPD, but the sentence that made that safe is the one that got removed

It previously continued: *"`asusd` automatically provides the standard D-Bus interface `net.hadess.PowerProfiles` so KDE Plasma and GNOME desktop power sliders continue to work natively without `power-profiles-daemon`."*

With the provider gone, a user who follows that instruction loses the desktop power slider, `powerprofilesctl` entirely, and every `HoldProfile` client — GameMode, Steam/Lutris wrappers, and the desktop environments' own transient boosts. The `asusctl.install` change compounds it: upgrades from 3.x no longer enable PPD either.

### Suggestion: document both resolutions

The contention is fixed by having *one* writer, not by having a *particular* writer — so there are two symmetric resolutions, and the manual currently documents only the one that costs the user functionality:

- **asusd owns it:** disable PPD (what the manual says today). asusd applies its configured EPP, fan curves and PPT limits, but the DE slider and all hold-based boosting are gone.
- **PPD owns it:** leave PPD running and stop asusd writing EPP — `platform_profile_linked_epp: false`, plus `change_platform_profile_on_ac: false` and `change_platform_profile_on_battery: false`. No contention, and the slider, `powerprofilesctl` and GameMode holds all keep working. Configuration only, no code.

That second one is what I've run since diagnosing #205 and it's been stable — PPD owns `net.hadess.PowerProfiles`, asusd never writes EPP, and the wrong-EPP symptom is simply absent. Given the provider is now out of scope, it might be worth the manual presenting both and letting the user choose which daemon owns the hardware, especially since the new warning points only at the first.

(For the record, I think dropping the provider was the right call — my `HoldProfile` note was about how much was left to do, not an argument for doing it.)

**Comment by qwertyman100** [comment]
Tested as promised. Built `3175cc16` (rustc 1.97.1), ran against a live `asusd` 6.3.10 with PPD 0.30 active.

**`rog-control-center`** — `RUST_LOG=debug cargo run -p rog-control-center`, first six lines of output:

```
[WARN  rog_control_center] ********************************************************************************
[WARN  rog_control_center] WARN: 'power-profiles-daemon' is active on system D-Bus!
[WARN  rog_control_center] Running power-profiles-daemon concurrently causes EPP and platform_profile contention.
[WARN  rog_control_center] To allow asusd to manage profiles, EPP, fan curves, and KDE/GNOME sliders natively, disable it:
[WARN  rog_control_center]   sudo systemctl disable --now power-profiles-daemon.service
[WARN  rog_control_center] ********************************************************************************
[INFO  rog_control_center] Running on G614FR, product: ROG Strix G16
```

**`asusctl`** — `asusctl profile get`:

```
WARN: 'power-profiles-daemon' is running concurrently with asusd!
This causes EPP and platform_profile contention.
Disable it via: 'sudo systemctl disable --now power-profiles-daemon.service'
Active profile: Balanced
```

Detection works, fires first thing, correctly identifies the name owner. Nothing was mutated by either run (config file md5 unchanged, `platform_profile`/EPP unchanged, PPD kept the same bus connection). ✅ **Works for me** — this would have saved me the investigation in #205.

**One observation from the test, which ties into my previous comment.** My machine is deliberately configured the other way round — PPD owns the hardware and asusd is configured not to write EPP:

```
platform_profile_linked_epp:        false
change_platform_profile_on_ac:      false
change_platform_profile_on_battery: false
```

In that configuration there is **no contention** — asusd never writes EPP, so nothing races — yet the warning fires anyway, on every single `asusctl profile` invocation including read-only ones like `get`. So as written it's a false positive for anyone who resolved #205 by making PPD the sole writer rather than by disabling it.

Cheap fix if you think it's worth it: gate the warning on asusd actually being configured to write, i.e. only warn when PPD is present **and** (`platform_profile_linked_epp || change_platform_profile_on_ac || change_platform_profile_on_battery`). Then it fires exactly when the two daemons will genuinely contend, which also makes the message strictly more trustworthy.

Not a blocker either way — I'd still merge this as-is.

File: distro-packaging/asusctl.install
Comment (Ghoul4500) [inline distro-packaging/asusctl.install:1]: The changes to this file seem unrelated to this PR

File: MANUAL.md [hunk 1]
```diff
@@ -71,7 +71,15 @@ These options are not written to the config file as they are stored in efivars.
...
-asusctl can support setting a power profile via platform_profile drivers. This requires [power-profiles-daemon](https://gitlab.freedesktop.org/hadess/power-profiles-daemon) v0.10.0 minimum. It also requires the kernel patch for platform_profile support to be applied form [here](https://lkml.org/lkml/2021/8/18/1022) - this patch is merged to 5.15 kernel upstream.
+asusctl supports setting power profiles via ACPI `platform_profile` drivers.
+
+> [!IMPORTANT]
+> **Compatibility with `power-profiles-daemon`**:
+> Do **NOT** run `power-profiles-daemon` alongside `asusd`. Running both services concurrently causes race conditions and contention over `/sys/firmware/acpi/platform_profile` and CPU EPP preferences.
+> 
+> To allow `asusd` to manage platform profiles, EPP, custom fan curves, and PPT limits cleanly, disable `power-profiles-daemon`:
+> ```bash
+> sudo systemctl disable --now power-profiles-daemon.service
```
Ghoul4500: closing ``` missing

asusctl/src/main.rs
```diff
@@ -804,6 +804,18 @@ fn handle_throttle_profile(
...
+        if let Ok(ppd_name) = zbus::names::BusName::try_from("net.hadess.PowerProfiles") {
+            if dbus_proxy.name_has_owner(ppd_name).unwrap_or(false) {
+                eprintln!("WARN: 'power-profiles-daemon' is running concurrently with asusd!");
+                eprintln!("This causes EPP and platform_profile contention.");
+                eprintln!(
+                    "Disable it via: 'sudo systemctl disable --now power-profiles-daemon.service'"
+                );
+            }
+        }
+    }
```
Ghoul4500: i feel this kind of placement will cause some spam. IMO, the readme alone is enough and the rest of these warnings can be dropped

Ghoul4500 (changes requested):
I'm mostly happy with this. but fold it into just 2 commits instead of these 3 one for readme and one for warnings, and then improve the wording to say any power profiles daemon is running, not necessarily ppd. There is one part that does this somewhat well. Consider disabling all EPP related features if any such daemon is running, since this would currently warn and proceed to conflict

Ghoul4500:
> I'm mostly happy with this. but fold it into just 2 commits instead of these 3 one for readme and one for warnings, and then improve the wording to say any power profiles daemon is running, not necessarily ppd. There is one part that does this somewhat well. Consider disabling all EPP related features if any such daemon is running, since this would currently warn and proceed to conflict

I think I'm to blame on this one for being a bit vague. I meant in rogcc and asusctl CLI we can disable changing EPP settings if PPD is detected, and give some feedback to user side. But at the same time, it does feel like more work than it's worth.

You can consider it but if it's a big change, I would rather just keep the warning in readme and drop all code change for now and think of a better permanent solution in time. I need to take a look at the EPP changes mario has been working on for that

Ghoul4500 (approved):
This looks good. Thank you for the contribution

PR #228, opened 2026-07-28
chore: distro packaging cleanup

scardracs:
# Description

This PR reorganizes and modernizes the packaging structure and GitHub Actions release workflow for `asusctl`:
- **Legacy Cleanup**: Cleaned up legacy 3.x upgrade migration logic, outdated 2021 kernel patch warnings, and automatic service toggles from `asusctl.install`, as well as 2020 package conflicts from `PKGBUILD`.

Fixes N/A

- **ASUS Laptop Model:** N/A (Packaging / CI Refactor)
- **Linux Distribution:** Arch Linux
- **Kernel Version:** N/A

# Verification and testing:


.github/workflows/release.yml
```diff
@@ -31,42 +31,20 @@ jobs:
...
+                  cp .github/packaging/PKGBUILD .
+                  cp .github/packaging/asusctl.install .
                   chown -R builder:builder .
                   sudo -u builder \
                       MAKEFLAGS="-j$(nproc)" \
                       CARGO_BUILD_JOBS="$(nproc)" \
                       CI_BUILD=1 \
                       _gitref="$_gitref" \
-                      pkgrel="$PKGREL" \
+                      pkgrel="${{ github.run_attempt }}" \
```
scardracs: @ayush123460 this way it can automatically understand if we are, for e.g., making v6.3.10, 6.3.10-1, etc without the need of a bucket server

scardracs:
For once I've made a PR with more line removed than added

.github/workflows/release.yml
ayush123460: @scardracs the pkgrel is actually used for the case where same version of the "source" is used to build, but packaging update was made. You can find more info on that here - https://man.archlinux.org/man/alpm-pkgrel.7.en

I'd prefer if we leave the pkgrel portion alone, why not rename the built files in the "if not release upload as build asset" step?

ayush123460:
The PKGBUILD is supposed to be usable outside of CI Builds by design, so moving it would defeat the purpose. Also I believe Terra was shipping `asusctl` and they may require the spec?

scardracs:
Ok I see this more an hassle than helpful. Closed ;)

Owen-sz:
> The PKGBUILD is supposed to be usable outside of CI Builds by design, so moving it would defeat the purpose. Also I believe Terra was shipping `asusctl` and they may require the spec?

We (Terra) don't require the spec here but I believe it was used to generate release artifacts.

scardracs:
I've killed the PR but I seriously think we need some deep cleanups of these legacies

Owen-sz:
Seems we do, you could probably reopen and just undo the removal of the PKGBUIKD/.spec?

scardracs:
> Seems we do, you could probably reopen and just undo the removal of the PKGBUIKD/.spec?

I'll take a good look at it tomorrow

Owen-sz:
> > Seems we do, you could probably reopen and just undo the removal of the PKGBUIKD/.spec?
> 
> I'll take a good look at it tomorrow 

Sounds good, I can help tomorrow too just let me know.

.github/workflows/release.yml
scardracs: Ok I leave it alone ;)

scardracs:
@ayush123460 @Owen-sz lemme know if it sits right with you now

distro-packaging/PKGBUILD
```diff
@@ -37,7 +30,9 @@ pkgver() {
 
 prepare() {
 	cd "$_workdir"
-	git checkout "$_gitref"
+	if [ "$_gitref" != "HEAD" ]; then
```
ayush123460, line 33: Why did you add this? Did the CI build fail when triggered manually?

distro-packaging/PKGBUILD
```diff
@@ -46,33 +41,32 @@ build() {
...
 	install="asusctl.install"
 
 	cd "$_workdir"
 	make DESTDIR="$pkgdir" install
-	rm -rf $pkgdir/usr/bin/rog-control-center
-	rm -rf $pkgdir/usr/share/rog-gui
-	rm -rf $pkgdir/usr/share/icons
-	rm -rf $pkgdir/usr/share/applications
-	rm -rf $pkgdir/usr/share/metainfo
+	rm -rf "$pkgdir"/usr/bin/rog-control-center
```
ayush123460, line 49: This and the other _package section to add quotes seems like it was added by mistake?

scardracs: I've added them so, if in the future we use a name like ABC DEF instead of ABC-DEF or ABC_DEF, it is still recognised as a valid name

ayush123460:
The rest are good. @Owen-sz if Terra doesn't need the spec file, is it fine to remove it from here? Pacman only needs the `PKGBUILD` and the `.install` files

Edit: LGTM other than the spec question

distro-packaging/PKGBUILD
scardracs: In `release.yml`, `_gitref` gets set to either the release tag (on `release`) or the explicit commit SHA from `$(git rev-parse HEAD)` (on `workflow_dispatch`), so `git checkout "$_gitref"` executes in both CI scenarios.

The `if [ "$_gitref" != "HEAD" ]` check in `PKGBUILD` is strictly for **local builds outside CI** (`makepkg`):
1. When running `makepkg` locally without setting `_gitref`, it defaults to `"HEAD"`. Skipping `git checkout HEAD` prevents detaching `HEAD` from the developer's local branch.
2. If a developer explicitly passes a target ref (e.g. `_gitref=v6.0.0 makepkg`), it correctly checks out that ref.

distro-packaging/PKGBUILD
ayush123460: Understood, thanks

distro-packaging/PKGBUILD
ayush123460: Got it, thanks for the explanation!

Owen-sz:
Yeah, I've confirmed we can remove the .spec.

scardracs:
@ayush123460 @Owen-sz done

Owen-sz (approved):
LGTM now

Ghoul4500 (approved):

PR #237, opened 2026-07-29
refactor(rog-platform): clean up CPU error messages and remove legacy TODO comments

scardracs:
## Description
This PR cleans up CPU error messages and removes obsolete inline comments in `rog-platform/src/cpu.rs`.
### Changes Included:
- Standardized error messages from `"No CPU's?"` to `"No CPUs found"`.
- Removed outdated `// TODO: check cpu are sync` comments.
## Verification
All workspace checks passed cleanly:
- `cargo check --all-targets`
- `cargo test --all`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo cranky`
- `cargo fmt --all -- --check`

### DIFF
--- a/rog-platform/src/cpu.rs
+++ b/rog-platform/src/cpu.rs
@@ -86,9 +86,8 @@ impl CPUControl {
         if let Some(path) = self.paths.first() {
             let s = read_attr_string(&to_device(path)?, ATTR_GOVERNOR)?;
             Ok(s.as_str().into())
-            // TODO: check cpu are sync
         } else {
-            Err(PlatformError::CPU("No CPU's?".to_string()))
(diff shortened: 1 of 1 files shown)


Ghoul4500 (changes requested):
I'm not understanding what this is or what the point of this PR is. Isn't this just dead code?

scardracs:
> I'm not understanding what this is or what the point of this PR is. Isn't this just dead code?

I meant for it to make a single file that check for all the features instead of spread them in different files. If you think it's not worth the time I can drop it

EDIT

In the end, I decided to proceed in the opposite direction to what was said

Ghoul4500 (approved):

PR #249, opened 2026-07-31
refactor: remove mocking module and associated feature flag

scardracs:


scardracs:
Ghoul in the end it was dead code so I decided to remove it completely

Ghoul4500 (approved):
Hmm, this does indeed seem to be dead code. But I wonder what luke was trying to do when he originally added this. I think it became obsolete later but some parts may be salvageable and/or usable. I'm marking this approved for me to merge later but in the meantime you can try to find what this was originally used for, and if it really has no business being here, I will merge this commit tonight

scardracs:
> Hmm, this does indeed seem to be dead code. But I wonder what luke was trying to do when he originally added this. I think it became obsolete later but some parts may be salvageable and/or usable. I'm marking this approved for me to merge later but in the meantime you can try to find what this was originally used for, and if it really has no business being here, I will merge this commit tonight

I already took a look at it: it was thought as a mock to be able to start asusctl on non-asus laptop to program/test out things. Honestly, I don't think someone would run asusctl on a non-asus laptop nowadays, also because most of the things are not working as was thought back in the days

Ghoul4500:
> I already took a look at it: it was thought as a mock to be able to start asusctl on non-asus laptop to program/test out things. Honestly, I don't think someone would run asusctl on a non-asus laptop nowadays, also because most of the things are not working as was thought back in the days

Looks like that is what it was made for. LGTM then

PR #252, opened 2026-08-01
[UI redesign 1/3] rog-control-center: build .mo from .po at build time

NB-Group:
Part of the UI redesign stack (1/3) — split out for easier review:

1. **this** — build .mo from .po at build time
2. #253 — fan sensor sentinel (missing vs stopped)
3. #230 — AC-style frontend redesign

The three PRs touch disjoint files and can be reviewed / merged independently in any order.

---

Add a `msgfmt` step in `build.rs` that compiles each `.po` into a `.mo` next to it, so translations rebuild from source instead of carrying stale binaries. Drop the checked-in `.mo` files and gitignore them.

PR #253, opened 2026-08-01
[UI redesign 2/3] rog-platform: fan sensor sentinel (missing vs stopped)

NB-Group:
Part of the UI redesign stack (2/3) — split out for easier review:

1. #252 — build .mo from .po at build time
2. **this** — fan sensor sentinel
3. #230 — AC-style frontend redesign

The three PRs touch disjoint files and can be reviewed / merged independently in any order.

---

Return `-1` when a fan sensor is absent and `0` when the fan is genuinely stopped, so consumers can tell "no sensor" apart from "idle 0 RPM" instead of treating both as zero. The UI redesign (#230) uses this sentinel to hide the GPU/mid fan rows on devices without those sensors.

### DIFF
--- a/rog-platform/src/platform.rs
+++ b/rog-platform/src/platform.rs
@@ -325,22 +325,25 @@ pub enum Properties {
 }
 
 pub fn get_fan_rpms() -> (i32, i32, i32) {
-    let mut cpu = 0;
-    let mut gpu = 0;
-    let mut mid = 0;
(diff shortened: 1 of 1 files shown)


NB-Group:
Merged into #230 per review direction (its fan sentinel commit now lives there), with the UI adapted for the new -1/0 semantics: missing sensors (-1) hide their row, a genuine 0 still shows as 0 RPM.

PR #257, opened 2026-08-03
asusd: open scsi_generic /dev/sgN for SCSI Aura devices

NB-Group:
SCSI Aura devices (e.g. ROG STRIX Arion, `0b05:1932`) showed up on D-Bus and `asusctl scsi` reported success, but no traffic ever reached the enclosure. The LEDs never changed.

`write_effect()` opened the block node (`/dev/sdX`) and issued `SG_IO` with the vendor CDB. SG_IO with a vendor command on a block node requires `CAP_SYS_RAWIO`, but the hardened asusd unit strips every capability (empty `CapabilityBoundingSet`, `NoNewPrivileges=true`), so every ioctl returned `EPERM`. `write_effect()` then did `perform(task).ok()` and threw the error away, so asusd reported success while zero commands hit the wire. strace showed `ioctl(N, SG_IO, ...) = -1 EPERM` sixteen times per write, and usbmon showed only baseline usb-storage traffic.

Fix: resolve the `scsi_generic` node (`/dev/sgN`) backing the block device and open that instead. The sg driver gates access at `open()` through the file permissions, so SG_IO works with no capabilities, the same path sg3_utils and OpenRGB use. I also stopped swallowing the ioctl error in `write_effect` so this can't go silent again.

Verified on an Arion: 16/16 SG_IO return 0, the ENE vendor CDBs (mode `0x8021`, colour registers `0x8160+`, apply `0x80a0`) show up in usbmon, and the LEDs change.

Two files, no kernel changes. This is the working userspace reference for the SCSI device-handler driver we discussed.

On hotplug the sg node can lag the block node by a moment, so init retries briefly before falling back to the block device; at startup the node already exists, so the first attempt hits with no delay.

### DIFF
--- a/asusd/src/aura_manager.rs
+++ b/asusd/src/aura_manager.rs
@@ -276,6 +276,31 @@ impl DeviceManager {
         Ok(devices)
     }
 
+    /// Resolve the `/dev/sgN` (scsi_generic) node backing a block device.
+    ///
+    /// Walks up from the block device to its owning scsi_device and reads the
+    /// `scsi_generic/sgN` child. Works for whole-disk (`/dev/sda`) and
+    /// partition (`/dev/sda1`) nodes alike, since the scsi_device is a common
+    /// ancestor. Returns None if no sg node exists (e.g. the `sg` module is
+    /// not loaded).
+    fn sg_node_for_block(device: &Device) -> Option<String> {
+        let mut current = device.parent();
+        while let Some(d) = current {
+            if let Ok(entries) = std::fs::read_dir(d.syspath().join("scsi_generic")) {
+                for entry in entries.flatten() {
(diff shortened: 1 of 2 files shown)


scardracs:
BTW, we own sg.rs too if you want to take a look into it too. The one on crates.io is not supported anymore, like the one made by luke back in 2025, so I took the liberty to pick his version up and import it directly into code.

NB-Group:
Nice, took a look. sg.rs sidesteps the cmd_len trap I hit in the kernel driver: it sets cmd_len from the slice length, so the 16-byte vendor CDB works as-is. The kernel sizes the CDB by opcode, 0xec maps to 10, cdb[13] (the ENE data-length byte) drops silently, and I had to hand-build the request there to force 16.

Ghoul4500 (approved):

PR #268, opened 2026-08-04
feat: add GU502GV aura support

MasterDash5:
adds aura_support.ron for GU502GV. Identical to GU502GU other than increased storage and RTX 2060 over GTX 1660.


Ghoul4500 (approved):

PR #271, opened 2026-08-05
Add Aura support for ROG Strix G16 G614PM

sensiedi:
G614PM shares the same Aura hardware ID (19b6) as G614FM but had no
  device_name entry in aura_support.ron, causing asusd to fall back to a
  default profile with no Lightbar/Logo zone support. Duplicated the
  G614FM entry under G614PM's name.

  Tested on hardware: keyboard, lightbar, and logo zones all respond
  correctly to basic effects (static, rainbow-cycle, etc.) in sync.


Ghoul4500 (approved):

PR #274, opened 2026-08-07
fix(rog-platform): select the dGPU by PCI vendor in telemetry fallbacks

NB-Group:
When NVML isn't available, get_gpu_usage_pct and get_gpu_temp scanned sysfs for the first node that matched, without checking which GPU it came from. On a hybrid laptop with an AMD APU alongside the NVIDIA dGPU, that first match is the APU, so the values shown as dGPU were really the integrated GPU.

This adds two helpers on top of the existing Device::find() (which already picks the dGPU by PCI vendor and whether it drives a display) and points both fallbacks at the dGPU's own drm card and hwmon directories. The iGPU helpers already filtered by vendor; the dGPU ones now match that.

On a box where NVML fails and the closed NVIDIA driver exposes no sysfs nodes, these still read nothing. That's a driver or userspace mismatch, not something the code can work around.

### DIFF
--- a/rog-platform/src/gpu_pci.rs
+++ b/rog-platform/src/gpu_pci.rs
@@ -618,6 +618,40 @@ pub fn get_gpu_names() -> (String, String) {
     )
 }
 
+/// Identify the dGPU once via `Device::find()` (PCI vendor + connected-panel
+/// check) and return its DRM card directory, e.g.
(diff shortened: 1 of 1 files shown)


Ghoul4500:
Superseded by #294

PR #277, opened 2026-08-07
Add Debian packaging

htrex:
This adds a `debian/` directory that builds two binary packages:

- **asusctl** — the `asusctl` CLI plus the `asusd`, `asusd-user` and `asus-shutdown` daemons, the udev rules, the D-Bus policy and the systemd units
- **rog-control-center** — the GUI, depending on the above

### Where it comes from

It started as a personal need rather than a plan to package asusctl for anyone else. I run it on Ubuntu on a ProArt P16 and wanted it installed as a package, so that upgrades, file ownership and removal behave like everything else on the system instead of being a `make install` I have to remember. It has been in daily use for a few months, published in a personal PPA:

https://launchpad.net/~hantarex/+archive/ubuntu/asusctl

The 6.3.11 build currently published there was produced with exactly this `debian/` directory. Nothing in it is specific to that PPA: the version suffixes (`+ds1`, `~series1`) are added by my own build scripts and are not part of what is proposed here. I'm opening this because the packaging turned out to be small and generic enough to be worth offering, not because anything is missing from the project.

### How it builds

Two modes, selected automatically by whether a vendored crates tarball sits next to the sources:

- **Offline from a vendored tarball** — the mode for anything you intend to ship, and what Launchpad builders and Debian buildds require, having no network during the build. The tarball fixes the dependency versions once, so a source package that built yesterday still builds today. `debian/README.Debian` documents how to produce it.
- **From a git checkout** — with no tarball present cargo resolves and downloads as usual. Convenient while working on the packaging, but not reproducible.

Both were exercised for this PR. The vendored mode built on a Launchpad builder (`Successfully built`), and the resulting packages are what I am running while writing this.

### Notes that apply regardless of this PR

Three things came up while getting the offline build to work, and they may be useful to know even if the packaging itself is not wanted:

- **`Cargo.lock` is gitignored**, so a tarball produced with `git archive` from a release tag carries no lock file, and `cargo build --locked` then fails with *"requires a lock file to be present first before it can be used against vendored source code"*. The packaging works around it by shipping the lock inside the vendored tarball, next to the crates it describes. Related: the `--locked` line in the `Makefile` is currently commented out with a note that the lock needs updating.

  This is more than a packaging inconvenience, and today it bites an ordinary build too. Without a lock, a fresh resolve currently picks `zune-core 0.5.2`, published on 2026-08-07, and `zune-jpeg 0.5.15` no longer compiles against it:

  ```
  error: macro expansion ends with an incomplete expression: expected expression
     --> zune-jpeg-0.5.15/src/mcu_prog.rs:463:17
  463 |   warn!("RST marker was not found, where expected, image may be garbled")
      |   ^ expected expression
  ```

  `warn!` moved from expanding to an expression to expanding to a statement, and `zune-jpeg` uses it as a block's trailing expression. `zune-jpeg 0.5.16-rc1` fixes it and was published 42 seconds later, but it is a pre-release, so cargo will not pick it.

  The consequence is that right now `make build`, and `distro-packaging/PKGBUILD` which calls it, fail from a clean tree — this is not specific to the packaging, it is what an unpinned resolve gets today. A build pinned to `zune-core 0.5.1` is fine, which is exactly what the vendored tarball does: it fixes the versions at the moment the release is prepared, and a breaking change published afterwards cannot reach back and break it. Committing Cargo.lock would give the same protection to everyone building from git — the usual Cargo guidance is that applications commit the lock and libraries do not.
- **`.cargo/config.toml`** routes linking through `.cargo/linker-wrapper.sh`, which picks mold or lld when installed. A distribution build wants a predictable result, so `debian/rules` overrides the linker to `cc` rather than depending on what the build host happens to have.
- **`rust-toolchain.toml`** pins `stable` and lists components. A rustup shim in `PATH` obeys it, so both `cargo` and `RUSTC` have to be pointed at the distribution toolchain explicitly, otherwise a build on a developer machine tries to install toolchain components mid-build.

### Deliberate choices

- No hand-written maintainer scripts. `asusd.service` has no `[Install]` section and is activated by udev through `SYSTEMD_WANTS` on `asus-nb-wmi`; debhelper's generated scripts already handle both units correctly, including stop on removal and `daemon-reload`.
- `Architecture: amd64` only.
- `Standards-Version: 4.7.4`, `debhelper-compat (= 13)`, `Rules-Requires-Root: no`, hardening enabled.
- lintian is clean apart from `no-manual-page`: the project ships no man pages, and inventing them here seemed worse than leaving the warning visible.

### If you'd rather shape it differently

The `Maintainer:` field and the changelog currently name me, which makes sense for a `debian/` maintained outside the tree and much less sense for one living here — happy to change both, or to move the directory, split it, or adjust anything else. Equally happy for this to be turned down; the notes above stand on their own either way.


htrex:
One applied, one not.

**Vendored source map** — applied in 604d01c7. The hand-written entries were in fact already wrong: `sg-rs` was listed but is not a dependency of the 6.3.11 tag nor of main, where slint is the only git source. It survived the cleanup that dropped supergfxctl, which is the failure mode you describe. `cargo vendor`'s own map now ships in the tarball and only its `directory` path is rewritten at build time; the build stops with an explicit message if the tarball predates the change, rather than silently falling back. Checked with an offline build.

**`--no-start`** — not applying this one. `dh_installsystemd` does add every installed unit to its list, but `deb-systemd-invoke` filters at runtime:

```perl
# If the job is disabled and is not currently running, the job is not started or restarted.
# We don't autostart static units either.
} elsif (!$unit_installed && !$unit_active && $action eq "restart") {
    print STDERR "$unit is a disabled or a static unit not running, not starting it.\n";
```

`systemctl is-enabled asusd.service` reports `static`, so on a fresh install — where udev has not activated it — the unit is skipped rather than started. What the restart does cover is the upgrade case: the daemon is already running, and without it the old binary stays in memory while rog-control-center is upgraded alongside. `--no-start` suppresses both, so it would give up that for preventing something that does not occur.

Ghoul4500:
After some discussion with @ayush123460 we have decided that the ublue tap is the best way we have debian packaging as of right now (also because of other issues regarding other packaging methods such as rustc versions). If you want to provide a manual installation guide for Debian, we can consider adding this when we have mdbook set up. Thank you for the effort you put into this. If you want the manual installation guide, you can follow up with us on discord and we'll let you know when mdbook is set up and ready

PR #279, opened 2026-08-08
fix(rog-control-center): app window default size increased to better …

mykola-shevchenko-dev:
# Description

Increases default `rog-control-center` window size to better fit content.
Before:
before
After:
after

- **ASUS Laptop Model:** ROG Strix SCAR 18 G835LX
- **Linux Distribution:** CachyOS (KDE Plasma)
- **Kernel Version:** 7.1.6-1-cachyos


Ghoul4500 (approved):

PR #280, opened 2026-08-08
fix: platform profile handling when quiet/low-power is missing

Boogie61:
My laptop (ROG Zephyrus G16 GU606AX, kernel 7.1.6) only has two profiles:

```
$ cat /sys/firmware/acpi/platform_profile_choices
balanced performance
```

Three things break because of that.

**Cycling gets stuck.** `PlatformProfile::next()` only checks `choices` for LowPower, so from Performance it returns Quiet even when Quiet doesn't exist:

```
$ asusctl profile next
Error: RogPlatform: platform_profile: (quiet) not supported
```

Once you're on Performance you can't get back to Balanced. Now it filters the cycle by what the kernel actually reports. Added tests for the two- and three-profile cases.

**EPP changes even when the profile doesn't.** Both `next_platform_profile()` and `set_platform_profile()` set EPP first, then bail out if the profile write fails, leaving EPP where it shouldn't be. After one `asusctl profile next`:

```
platform_profile   performance   (unchanged, write failed)
EPP                power         (changed anyway)
```

So the CPU sits on the most power-saving preference while the profile still says Performance. Now the profile goes first and EPP only follows if it worked.

**Battery switching never happens.** The existing Quiet fallback needs LowPower to be present, which it isn't here, so the default `platform_profile_on_battery: Quiet` just fails every time. Unplugged:

```
             AC            battery
profile      performance   performance   (no change)
throttle     1             1             (no change)
```

The laptop stays at full performance on battery. Now any unavailable profile falls back to the closest one that exists. Custom is left alone since it never shows up in `choices`.

Tested on the GU606AX. `cargo test -p rog_platform` passes, `cargo build -p asusd` is clean.

### DIFF
--- a/asusd/src/ctrl_platform.rs
+++ b/asusd/src/ctrl_platform.rs
@@ -273,6 +273,35 @@ impl CtrlPlatform {
         }
     }
 
+    /// Map a configured profile onto one the kernel actually offers.
+    ///
+    /// The default config uses Quiet, 6.11+ renamed it to LowPower, and some
+    /// laptops have neither. Returns `None` when `configured` is usable as-is.
+    /// Custom is userspace-only and never in choices, so it is left alone.
+    ///
+    /// This only computes: nothing is written here, so a profile that turns
+    /// out to be unappliable is never persisted.
+    fn normalize_profile(&self, configured: PlatformProfile) -> Option<PlatformProfile> {
+        if configured == PlatformProfile::Custom {
+            return None;
+        }
+
+        let choices = self.platform.get_platform_profile_choices().ok()?;
+        if choices.is_empty() || choices.contains(&configured) {
+            return None;
+        }
+
+        // Prefer low-power, else the least aggressive one available.
+        [
+            PlatformProfile::LowPower,
+            PlatformProfile::Quiet,
+            PlatformProfile::Balanced,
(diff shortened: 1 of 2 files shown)


Boogie61:
Fixed in 647e726: `normalize_profile()` now only computes, and the config write moved to after `set_platform_profile()` succeeds. The AC and battery setters had the same problem plus their own narrower Quiet->LowPower fallback, so they now share the same path.

The `Custom` finding is real but predates this branch — the old code passed `Custom` straight through as well, since the normalization only ever matched `Quiet`. Leaving `Custom` untouched here is deliberate so a user setting is not clobbered; it just makes the existing gap easier to see. Fixing it means deciding what `Custom` should mean at the platform level, either skipping the kernel write and applying only its EPP, or picking an underlying profile. That is a separate change, out of scope for this branch.

Ghoul4500:
I believe this is a symptom of a bigger issue that asusd doesn't infer the available profiles from what the kernel allows and instead uses the hardcoded enum. If possible, could you make it so this reads the profiles and uses them instead of juggling choices?

Ghoul4500:
Please reopen or make a new PR as described above. It is majorly different from what is implemented here (and because author is unresponsive), so I am closing this

PR #289, opened 2026-08-12
fix(asusd): bind GX651AR AURA button listener to matching HID device

Cymirk:
The ASUS ROG Zephyrus Duo GX651AR detachable keyboard exposes its
Aura button through the same HID interface used for Aura LED control.

On the GX651AR, the AURA button generates an input report on the
0b05:1ce6 USB keyboard's Aura HID interface (interface 1.2). The
listener previously discovered this HID node independently, which
could allow the listener to become associated with the wrong Aura
device when multiple Aura devices were present.

This change binds the AURA button listener to the specific HID device
being initialized. The listener is only started when the current
device is the verified GX651AR Aura interface, and its already-resolved
hidraw device node is passed directly to the listener.

This keeps the button listener associated with the correct
`AuraZbus` instance and HID device, while avoiding a second independent
device scan.

The listener then uses the existing Aura mode-changing path to advance
through the supported Aura modes when the GX651AR AURA button is
pressed.

- **ASUS Laptop Model:** ROG Zephyrus Duo GX651AR
- **Keyboard:** ASUS 0b05:1ce6 detachable keyboard
- **Aura interface:** USB interface 1.2
- **Linux Distribution:** Fedora 44

## Verification and testing

- Self-reviewed the change
- `cargo clippy --all -- -D warnings`
- `cargo cranky`
- Built and installed `asusd` successfully
- Verified the GX651AR AURA button is detected
- Verified each button press advances exactly one Aura mode
- Verified mode cycling wraps correctly:
  `Static → Breathe → RainbowCycle → Pulse → Static`
- Verified Aura mode changes continue to work without requiring
  repeated button presses
- Verified the listener is bound to the matching GX651AR HID interface

### DIFF
--- a/asusd/src/aura_laptop/trait_impls.rs
+++ b/asusd/src/aura_laptop/trait_impls.rs
@@ -137,7 +137,9 @@ impl AuraZbus {
         if config.brightness == LedBrightness::Off {
             config.brightness = LedBrightness::Med;
         }
-        self.0.set_brightness(config.brightness.into()).await?;
+        if let Err(e) = self.0.set_brightness(config.brightness.into()).await {
+            log::warn!("Could not set keyboard backlight brightness: {e}");
+        }
         config.write();
         Ok(())
     }
--- a/asusd/src/aura_laptop/trait_impls.rs
+++ b/asusd/src/aura_laptop/trait_impls.rs
@@ -179,7 +179,9 @@ impl AuraZbus {
         if config.brightness == LedBrightness::Off {
             config.brightness = LedBrightness::Med;
         }
-        self.0.set_brightness(config.brightness.into()).await?;
+        if let Err(e) = self.0.set_brightness(config.brightness.into()).await {
+            log::warn!("Could not set keyboard backlight brightness: {e}");
+        }
         config.set_builtin(effect);
         config.write();
         Ok(())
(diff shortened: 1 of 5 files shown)


Ghoul4500:
Isn't the missing backlight control an issue in the kernel rather than something we need workarounds for in userspace? The warnings aren't necessarily wrong. Also this PR seems to do a lot more than that..

scardracs:
> Isn't the missing backlight control an issue in the kernel rather than something we need workarounds for in userspace? The warnings aren't necessarily wrong. Also this PR seems to do a lot more than that..

Yeah that should be kernel side rather than on asusctl

Cymirk:
sorry im going to rework on this

Ghoul4500:
> sorry im going to rework on this

You could open an issue with relevant logs and details. Since we already have ongoing keyboard backight work, this fix can be sent as part of my upcoming patches. Tag me in the issue you create and close this PR

Cymirk:
Thanks for the feedback. I reworked the PR and removed the keyboard backlight workaround entirely.

The PR is now limited to the GX651AR AURA button handling. The listener is bound to the already-identified GX651AR 0b05:1ce6 Aura HID interface rather than performing a separate HID scan, and it uses the existing Aura mode-setting path to cycle modes.

I also verified the button behavior on the GX651AR: each press advances exactly one mode through Static → Breathe → RainbowCycle → Pulse → Static.

This should leave the missing generic keyboard backlight control as a separate kernel-side issue.

PR #294, opened 2026-08-14
fix(gpu): overhaul GPU telemetry with passive zero-wakeup monitoring, udev optimizations, and remove NVML (#293)

### LINKED ISSUE #293: [Bug]: rog-control-center telemetry loop calls NVML unconditionally, resuming a runtime-suspended dGPU while the window is open
**Affected Component:** rog-control-center (GUI / Slint / System Tray) — root cause in rog-platform
**Version:** 6.3.11.r0.g4d8a45b3-1, and still present on `main` (46 commits past the 6.3.11 tag)
**Session:** Wayland (Hyprland)

## Issue Description

While the ROG Control Center **main window is open**, the telemetry loop calls
`nvml_wrapper::Nvml::init()` unconditionally every 2 seconds. `Nvml::init()` opens the
NVIDIA device nodes, which resumes a runtime-suspended dGPU. The card stays in `active`
(D0) for as long as the window is open, and returns to `suspended` once it is closed.

**This is not a duplicate of #208 / #231 / #234 — it is the half of #208 that was never fixed.**
#208 was closed by #212, which pauses the telemetry loop *when the window is hidden*. That
fixes RCC sitting in the background. The path that remains is the window being **visible**,
which is the normal way the GUI is used. The gating fix proposed in #234's plan — "gate NVML
access behind the existing passive `GfxPower` runtime-status read" — was never implemented.

Note that #208 already has a post-6.3.11 report of residual behaviour
([comment by @fl4shb1c, 2026-08-04](https://github.com/OpenGamingCollective/asusctl/issues/208#issuecomment-5181130989))
that received no reply while the issue was closed. @ayush123460 asked in that thread for a
fresh issue if the problem persisted on 6.3.11 — it does.

## Root cause

`rog-platform/src/gpu_pci.rs` — both stat helpers call NVML with no power-state guard:

```rust
pub fn get_gpu_temp() -> f32 {
    if let Ok(nvml) = nvml_wrapper::Nvml::init() {          // line 668, no GfxPower check
        if let Ok(device) = nvml.device_by_index(0) {
            if let Ok(temp) = device.temperature(TemperatureSensor::Gpu) { return temp as f32; }

pub fn get_gpu_usage_pct() -> f32 {
    if let Ok(nvml) = nvml_wrapper::Nvml::init() {          // line 696, same
```

`rog-control-center/src/ui/setup_system.rs` — the caller runs every 2 s while visible:

```rust
loop {
    let visible = app_state.try_lock().map(|s| *s == AppState::MainWindowOpen).unwrap_or(false);
    if !visible {                                            // #212's fix — only covers hidden
        tokio::time::sleep(Duration::from_secs(2)).await;
        continue;
    }
    ...
    let gpu_temp  = rog_platform::gpu_pci::get_gpu_temp();      // line 146 -> Nvml::init()
    let gpu_usage = rog_platform::gpu_pci::get_gpu_usage_pct(); // line 151 -> Nvml::init()
    ...
    tokio::time::sleep(Duration::from_secs(2)).await;
}
```

So each iteration performs **two** separate NVML init/shutdown cycles.

Both files are identical on `main` as of today (46 commits past the 6.3.11 tag `4d8a45b3`),
so this still reproduces on `main`.

The system tray is **not** implicated: `tray.rs` is driven by the sysfs status monitor in
`notify.rs` and never touches NVML.

## Steps to reproduce

1. Ensure the dGPU is runtime-suspended:
   `cat /sys/bus/pci/devices/0000:01:00.0/power/runtime_status` → `suspended`
2. Open ROG Control Center (window visible).
3. Watch `runtime_status` and `power/runtime_suspended_time`.
4. Close the window.

## Measurements

All of the sampling below reads only sysfs and `/proc/<pid>/fd`; it changes nothing. Reading
`runtime_status` does not resume the device — it matches @ayush123460's reply in #208, and is
why the same read is safe to use as the guard proposed at the end of this report.

**Before opening RCC** — card asleep, and it accumulates sleep normally:

```
power/runtime_status           suspended
power/runtime_suspended_time   40133651
power/runtime_active_time       1022236
```

**With the RCC window open.** Reading the two counters when the window was opened, and again
13.5 minutes later with it still open:

```
                              at open     13m31s later     delta
runtime_status                active      active
runtime_suspended_time        42697627    42697627         +0 ms
runtime_active_time            1044424     1855393    +810969 ms
```

`runtime_suspended_time` does not advance by a single millisecond: the card is held in D0 for
the entire time the window is open. A 40-sample 1 Hz log over the same period shows `active`
on every sample, with `runtime_active_time` growing ~1016 ms per second.

**Direct evidence of the NVML calls.** Scanning `/proc/<pid>/fd` at ~1855 Hz for 12 s while
the window was open (1 Hz sampling misses it — each NVML cycle is only ~70 ms):

```
scans: 22263 over 12.0 s
   768  /dev/nvidia0
   286  /dev/nvidiactl
   279  /dev/nvidia-caps/nvidia-cap2
     2  /usr/lib/libnvidia-ml.so.610.57.04   <- dlopen during Nvml::init()
     1  /proc/driver/nvidia/params
     1  /dev/nvidia-uvm
```

`/dev/nvidia0` is open in 3.4% of scans, in a repeating pattern consistent with the 2 s loop.

**After closing the window**, the card returns to `suspended` on its own. Counters read from
the moment the window was opened:

```
total time in D0 while the window was open   995566 ms  (16m 35s)   sleep gained: 0 ms
time asleep since the window was closed      211104 ms  (3m 31s)
```

**Control: polling sysfs at 1 Hz does not keep the card awake.** #208 contains a hypothesis
that reading `runtime_status` every ~1 s resets the kernel's `last_busy` timestamp and is
itself what blocks suspend. It does not. With RCC closed and the card asleep, 60 samples at
1 Hz:

```
samples not 'suspended':          0 / 60
runtime_suspended_time increase:  60310 ms over 60 s
```

The card slept through the entire polling run. This is why the guard suggested below is safe:
the read it depends on does not resume the device.

## Current behaviour

The dGPU is held in D0 for the entire time the RCC window is open. On battery this is a
large, continuous power cost on a laptop whose dGPU otherwise idles in D3cold — which in
practice means the GUI cannot be opened to check a fan curve or a charge limit without
waking the card.

## Expected behaviour

Opening the GUI should not resume a sleeping dGPU. When `get_runtime_status()` reports
anything other than `GfxPower::Active`, the NVML calls should be skipped and the GPU
temperature/usage fields should show an idle/unavailable placeholder.

## Suggested fix

1. **Gate NVML in `rog-platform`** (the fix #234's plan called primary, so every caller is
   covered): in `get_gpu_temp()` and `get_gpu_usage_pct()`, read the dGPU's
   `power/runtime_status` first — this is a passive sysfs read that does not wake the device —
   and return early unless the state is `Active`.
2. **Stop re-initialising NVML every iteration.** Even when the card *is* awake, the loop
   does `Nvml::init()` + drop twice per 2 s, which is what the `/proc/<pid>/fd` scan above
   shows. Holding a single handle for the process lifetime would avoid the repeated
   open/close of the device nodes.

## System details

- Distro: Arch Linux
- Kernel: 7.1.8-zen1-3-zen
- Desktop: Hyprland (Wayland)
- Model: ROG Strix G513RM (board G513RM)
- dGPU: NVIDIA GA106M [GeForce RTX 3060 Mobile / Max-Q], PCI `0000:01:00.0`
- NVIDIA driver: 610.57.04 (open kernel module)
- `/etc/modprobe.d`: `options nvidia NVreg_EnableS0ixPowerManagement=1 NVreg_DynamicPowerManagement=0x02`
- asusctl / rog-control-center: 6.3.11.r0.g4d8a45b3-1 (from the `[ogc]` repo)

scardracs:
# Description

This PR resolves periodic dGPU wakeups caused by NVML queries in `rog-control-center` and transitions GPU telemetry to a completely vendor-agnostic, zero-wakeup Linux sysfs/`hwmon`/DRM implementation.

### Key Changes:
- **Zero-Wakeup & Runtime PM Awareness**: Gated dGPU metrics (`temp` and `usage_pct`) behind PCI `runtime_status`. If the dGPU is `Suspended`, `AsusDisabled`, or `Unknown`, metrics immediately return default values (`0.0`) without querying hardware nodes or waking up the PCIe link.
- **Vendor-Agnostic Monitoring**: Replaced `nvml-wrapper` with standardized kernel `/sys/class/hwmon` (`temp1_input` in millidegrees Celsius) and DRM (`gpu_busy_percent`) interfaces, enabling seamless telemetry support for AMD Radeon, Intel, and NVIDIA GPUs.
- **Dependency Cleanup**: Completely removed `nvml-wrapper` from workspace dependencies across `Cargo.toml`, `rog-platform`, and `rog-control-center`.
- **Aggregated Telemetry & Performance**: Introduced `GpuTelemetry` and `get_gpu_telemetry()` in `rog-platform::gpu_pci` to fetch iGPU/dGPU metrics in a single udev pass, reducing 4 redundant PCI udev enumerations every 2 seconds to 1 in `rog-control-center`.
- **Refactoring & Cleanup**: Removed external `lspci` subprocess spawning in fallback paths in favor of udev/PCI database matching; fixed naming typos (`lscpi` -> `lspci`, `start_dpu_status_mon` -> `start_dgpu_status_mon`), and removed unused commented-out code.

Fixes #293

- **ASUS Laptop Model:** ASUS ROG G614PR
- **Kernel Version:** 7.2-rc7

# Verification and testing:

### DIFF
--- a/rog-platform/src/gpu_pci.rs
+++ b/rog-platform/src/gpu_pci.rs
@@ -155,6 +155,40 @@ pub fn is_display_class(pci_class: &str) -> bool {
     u32::from_str_radix(pci_class, 16).is_ok_and(|class| class >> 16 == 0x03)
 }
 
+fn read_hwmon_temp(dir: &Path) -> Option<f32> {
+    fs::read_to_string(dir.join("temp1_input"))
+        .ok()?
+        .trim()
+        .parse::<f32>()
+        .ok()
+        .map(|t| t / 1000.0)
+}
+
+fn read_drm_busy(dir: &Path) -> Option<f32> {
+    fs::read_to_string(dir.join("device/gpu_busy_percent"))
+        .or_else(|_| fs::read_to_string(dir.join("gpu_busy_percent")))
+        .ok()?
+        .trim()
+        .parse::<f32>()
+        .ok()
+}
+
+fn read_nvml_temp() -> Option<f32> {
+    let nvml = nvml_wrapper::Nvml::init().ok()?;
+    let device = nvml.device_by_index(0).ok()?;
+    let temp = device
+        .temperature(nvml_wrapper::enum_wrappers::device::TemperatureSensor::Gpu)
+        .ok()?;
+    Some(temp as f32)
+}
+
+fn read_nvml_usage() -> Option<f32> {
+    let nvml = nvml_wrapper::Nvml::init().ok()?;
+    let device = nvml.device_by_index(0).ok()?;
+    let rates = device.utilization_rates().ok()?;
+    Some(rates.gpu as f32)
+}
+
 // --- Device ---
 
 /// A PCI GPU device.
@@ -183,18 +217,8 @@ impl Device {
 
     /// Read a file underneath the sys object.
     fn read_file(path: PathBuf) -> Result<String> {
-        let path = path
-            .canonicalize()
-            .map_err(|e| PlatformError::Read(path.to_string_lossy().to_string(), e))?;
-        let mut data = String::new();
-        let mut file = fs::OpenOptions::new()
-            .read(true)
-            .open(&path)
-            .map_err(|e| PlatformError::Read(path.to_string_lossy().to_string(), e))?;
-        trace!("read_file: {file:?}");
-        file.read_to_string(&mut data)
-            .map_err(|e| PlatformError::Read(path.to_string_lossy().to_string(), e))?;
-        Ok(data)
+        fs::read_to_string(&path)
+            .map_err(|e| PlatformError::Read(path.to_string_lossy().to_string(), e))
     }
 
     /// Read the runtime power status from sysfs.
@@ -224,14 +248,10 @@ impl Device {
         }
 
         // 1. Direct hwmon directory under device path
-        let hwmon_dir = self.dev_path.join("hwmon");
-        if let Ok(entries) = fs::read_dir(&hwmon_dir) {
+        if let Ok(entries) = fs::read_dir(self.dev_path.join("hwmon")) {
             for entry in entries.flatten() {
-                let temp_path = entry.path().join("temp1_input");
-                if let Ok(temp_str) = fs::read_to_string(temp_path) {
-                    if let Ok(temp_val) = temp_str.trim().parse::<f32>() {
-                        return Some(temp_val / 1000.0);
-                    }
+                if let Some(temp) = read_hwmon_temp(&entry.path()) {
+                    return Some(temp);
                 }
             }
         }
@@ -246,11 +266,8 @@ impl Device {
                         || p.starts_with(&self.dev_path)
                 });
                 if is_match {
-                    let temp_path = path.join("temp1_input");
-                    if let Ok(temp_str) = fs::read_to_string(temp_path) {
-                        if let Ok(temp_val) = temp_str.trim().parse::<f32>() {
-                            return Some(temp_val / 1000.0);
-                        }
+                    if let Some(temp) = read_hwmon_temp(&path) {
+                        return Some(temp);
                     }
                 }
             }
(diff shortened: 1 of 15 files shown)


luytan:
if the dGPU is suspended, there will be no telemetry for it?

rog-platform/src/gpu_pci.rs
```diff
@@ -522,102 +583,62 @@ pub fn get_gpu_names() -> (String, String) {
...
-                    if let Ok(val) = val_str.trim().parse::<f32>() {
-                        return val;
-                    }
-                }
-            }
-        }
-    }
-    0.0
+    get_gpu_telemetry().dgpu_usage
 }
```
Ghoul4500: do we need all these wrapper functions when `get_gpu_telemetry` is public anyway?

Ghoul4500 (changes requested):

Ghoul4500 (approved):

PR #307, opened 2026-08-16
Add Aura support for ROG Strix G16 G614PP

deminearchiver:
# Description

Adds Aura support for ASUS ROG Strix G16 G614PP laptop.
Manually verified by comparing to Armoury Crate available Aura features on Windows.
The configuration works when loaded locally into ROG Control Center as well.
Manually verified that all other modes except the ones listed in the entry are unsupported.
This model doesn't seem to have "basic zones".
Extracted `product_id` value properly using `lsusb`.
The configuration is almost identical to G614PR, but I've decided to omit `basic_zones` because they seem to have no effect (I have no idea what they do when per-key lighting is supported).

Fixes # (issue)

- **ASUS Laptop Model:** ROG Strix G16 G614PP-S5063.
- **Linux Distribution:** EndeavourOS.
- **Kernel Version:** `6.18.44-1-lts`, `7.1.8-arch1-3`, and `7.2.0-rc7-1-ogc`.

# Verification and testing:

### DIFF
--- a/rog-aura/data/aura_support.ron
+++ b/rog-aura/data/aura_support.ron
@@ -450,6 +450,15 @@
         power_zones: [Keyboard, Lightbar, Logo],
     ),
     (
+        device_name: "G614PP",
+        product_id: "19b6",
+        layout_name: "g634j-per-key",
+        basic_modes: [Static, Breathe, RainbowCycle, RainbowWave, Pulse],
+        basic_zones: [],
+        advanced_type: PerKey,
+        power_zones: [Keyboard, Lightbar],
+    ),
+    (
         device_name: "G614PR",
         product_id: "",
         layout_name: "g614pr-per-key",


Ghoul4500 (approved):

scardracs:
Why did you use the 634 per key instead the 614 per key?

deminearchiver:
> Why did you use the 634 per key instead the 614 per key?

I've compared both layouts and it seemed to me that 634 was closer to the physical layout. It might've been an oversight on my side. Does the locale specified in the layout file matter? My laptop, for example, doesn't seem to have ISO keys (large enter, short lshift, etc.), so I thought that non-ISO would be the right layout choice.

scardracs:
> > Why did you use the 634 per key instead the 614 per key?
> 
> I've compared both layouts and it seemed to me that 634 was closer to the physical layout. It might've been an oversight on my side. Does the locale specified in the layout file matter? My laptop, for example, doesn't seem to have ISO keys (large enter, short lshift, etc.), so I thought that non-ISO would be the right layout choice.

I see. Not that it really matters as it is configure the LEDs layout rather than the buttons themselves

PR #316, opened 2026-08-18
feat(asusd): remember platform profile per source

estoult:
# Description

Previously, switching between AC and battery always applied the fixed platform_profile_on_ac/platform_profile_on_battery values, overwriting any profile manually chosen while on that power source.

Introducing new flag: remember_profile_for_current_source

Now, if the flag is true, the last profile manually selected on a given power source is remembered and reapplied the next time that source becomes active.

- Add remember_profile_for_current_source() to persist the active profile into platform_profile_on_ac/battery whenever it's changed via set_platform_profile or next_platform_profile.
- Add remember_platform_profile_per_source config flag (default false) to gate this behavior, with a matching DBus property/setter.
- Enabling the flag immediately captures the current profile for the active power source instead of waiting for the next manual change.
- Expose the new property in the rog-dbus Platform proxy.

- **ASUS Laptop Model:** ROG Zephyrus G14 GA403WR_GA403WR
- **Linux Distribution:** Fedora Linux 44
- **Kernel Version:** 7.1.8-200.fc44.x86_64 (64-bit)

# Verification and testing:

### DIFF
--- a/MANUAL.md
+++ b/MANUAL.md
@@ -93,6 +93,17 @@ asusctl supports setting power profiles via ACPI `platform_profile` drivers.
 >    platform_profile_linked_epp: false,
 >    ```
 
+By default, `asusd` always applies the configured
+`platform_profile_on_ac`/`platform_profile_on_battery` values whenever
+you switch power source. To instead have `asusd` remember the last
+profile you manually selected on each power source and reapply it the
+next time that source becomes active again, enable the following in
+`/etc/asusd/asusd.ron`:
+
+```ron
+remember_platform_profile_per_source: true,
+```
+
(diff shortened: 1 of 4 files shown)


Ghoul4500:
Hmm, this is a good change to bring but this conflicts with our plans of imitating armoury crate to show a manual mode instead of per profile settings. And then we will have custom profiles as well

It's a work in progress already so I think it's better to go that route. You can join us onndiscord if you'd like to contribute

estoult:
@Ghoul4500 It's okay, now I can see the ongoing design. I'll move this feature in a custom detached daemon. Thank you for exposing the DBus interfaces.
I'll keep an eye on the project development!

PR #336, opened 2026-08-22
fix(rog-platform): prevent dGPU wakeups by gating NVML on active state and filtering non-display functions

### LINKED ISSUE #302: [Bug]: dGPU still resumed every 2 s on 6.4.0 — #294's guard is bypassed via the dGPU's sibling audio function, and iGPU telemetry reports the dGPU
**Affected Component:** rog-control-center (GUI / Slint / System Tray) — root cause in rog-platform
**Version:** 6.4.0.r0.ge6c1469c-1, and identical on `main` (`main` is currently the 6.4.0 tag, 0 commits ahead)
**Session:** Wayland (Hyprland)

## Issue Description

This is a follow-up to #293, which #294 fixed and closed. The guard #294 added is correct and it
does fire — but there is a **second path into NVML that never reaches it**, so on 6.4.0 the dGPU is
still resumed roughly every 2 seconds while the RCC window is open.

The guard keys on the `Device::is_dgpu` flag. `Device::find()` also returns the dGPU's **sibling PCI
function** — on this machine `0000:01:00.1`, the NVIDIA HD Audio controller (`10DE:228E`, class
`040300`) — with `is_dgpu: false` but `pci_id: "10DE:228E"`. Every guard in `gpu_pci.rs` is written
`if self.is_dgpu && runtime_status != Active`, while the NVML fallback below it is written
`if self.pci_id.starts_with("10DE")`. A non-display NVIDIA function passes between the two: the
guard skips it because the flag is false, and the fallback claims it because the vendor matches.

`get_gpu_telemetry()` then treats that audio function as the *integrated* GPU, calls `get_temp()` and
`get_usage_pct()` on it, falls through to NVML, and `Nvml::init()` + `device_by_index(0)` resumes the
sleeping dGPU. This also produces a second, user-visible symptom: **the iGPU temperature and usage
shown in the GUI are the dGPU's NVML values**, not the integrated GPU's.

Improvement over 6.3.11 is real but small: the card now manages about 6% sleep with the window open,
where before it managed 0%.

### 1. A non-display sibling function enters the device list as a non-dGPU NVIDIA device

`rog-platform/src/gpu_pci.rs`, in `Device::find()`:

```rust
if dgpu || (!parent.is_empty() && sysname.contains(&parent)) {   // line 476
    if dgpu {
        info!("Found dgpu {id} at {:?}", device.sysname());
    } else {
        info!("Found additional device {id} at {:?}", device.sysname());
    }
    parent = get_parent(&device);
    devices.push(Self {
        dev_path: PathBuf::from(device.syspath()),
        is_dgpu: dgpu,                                            // false for 01:00.1
        pci_id: id.to_string(),                                   // but still "10DE:228E"
    });
}
```

There is no class check on the `sysname.contains(&parent)` branch, so the audio function of the GPU
is admitted alongside it.

### 2. The guard checks the flag, the fallback checks the vendor

```rust
fn probe_hwmon(&self, read: fn(&Path) -> Option<f32>, nvml: fn() -> Option<f32>) -> Option<f32> {
    if self.is_dgpu                                               // line 281 — false here, guard skipped
        && self.get_runtime_status().unwrap_or(GfxPower::Unknown) != GfxPower::Active
    {
        return None;
    }
    // ... 01:00.1 has no hwmon and no DRM node, so both lookups miss ...
    if self.pci_id.to_uppercase().starts_with(NVIDIA_PCI_VENDOR) { // line 315 — matches
        return nvml();                                            // -> Nvml::init(), device_by_index(0)
    }
    None
}
```

`get_usage_pct()` has the same shape: guard at line 333, NVML fallback at line 372.

### 3. `get_gpu_telemetry()` routes it through the iGPU branch

```rust
if let Ok(devices) = Device::find() {
    for device in devices {
        if device.is_dgpu() {
            if dgpu_active { /* correctly skipped while suspended */ }
        } else {
            telemetry.igpu_temp  = device.get_temp().unwrap_or(-1.0);      // 01:00.1 -> NVML -> dGPU
            telemetry.igpu_usage = device.get_usage_pct().unwrap_or(-1.0); // 01:00.1 -> NVML -> dGPU
        }
    }
}
```

### 4. The `break` ends enumeration before the real iGPU is ever seen

```rust
if !parent.is_empty() && !sysname.contains(&parent) {   // line 492
    break;
}
```

On this machine the dGPU is `0000:01:00.0` and the integrated Radeon 680M is `0000:05:00.0`. Once
`01:00.1` has been consumed, the next PCI device does not match `parent` and the scan stops — the
iGPU is never enumerated at all. This is why the iGPU fields have nothing left to report but the
NVML value picked up in step 2. `get_gpu_names()` uses a separate scan that filters on display class
and has no `break`, which is why the GUI shows the correct iGPU *name* next to the wrong numbers.

### 5. The NVML handle now persists for the process lifetime

```rust
fn nvml() -> Option<&'static nvml_wrapper::Nvml> {
    static NVML: OnceLock<nvml_wrapper::Nvml> = OnceLock::new();   // line 182
```

This is a good change in itself, but it means one unguarded call is permanent: after the first
`Nvml::init()` the NVIDIA device nodes stay open for as long as RCC runs. In #293 `/dev/nvidia0` was
open in 3.4% of `/proc/<pid>/fd` scans; on 6.4.0 it is open in 100% of them.

## Steps to reproduce

1. Hybrid mode, dGPU runtime-suspended:
   `cat /sys/bus/pci/devices/<dgpu>/power/runtime_status` → `suspended`
2. Open ROG Control Center and leave the window visible.
3. Watch `power/runtime_status` and `power/runtime_suspended_time`.
4. Compare the "iGPU" temperature in the GUI against the integrated GPU's own hwmon
   (`/sys/bus/pci/devices/<igpu>/hwmon/hwmon*/temp1_input`).

Reproduces on any machine where the dGPU's audio function is enumerated (i.e. essentially every
Optimus laptop) — nothing here is specific to this board.

## Measurements

All sampling reads only sysfs and `/proc/<pid>/fd`.

**The card is still held awake.** 60 s with the window open, against a 60 s control with RCC closed:

```
window open      sleep gained:  3937 ms / 60 s   (6.0%)   active gained: 62148 ms
RCC closed       sleep gained: 60079 ms / 60 s (100.0%)   active gained:     0 ms
```

**NVML is entered while the card is still asleep.** `/proc/<pid>/fd` scanned from process start
(206140 scans in 30 s), timestamps relative to launch:

```
0.000  runtime_status = suspended
0.404  /dev/nvidiactl opened          <- card still 'suspended'; the guard never saw this call
0.674  runtime_status -> resuming
2.152  runtime_status -> active
2.153  /dev/nvidia0, /dev/nvidia-caps/nvidia-cap2, /dev/nvidia-uvm opened
22.357 runtime_status -> suspending
22.973 runtime_status -> suspended
24.625 runtime_status -> resuming    <- next telemetry tick wakes it again
26.103 runtime_status -> active
```

**It is NVML, not the renderer.** The only NVIDIA library mapped into the process is
`/usr/lib/libnvidia-ml.so.610.57.04`; the EGL vendor actually loaded is `libEGL_mesa.so`, with no
`libGLX_nvidia`, no `libEGL_nvidia` and no Vulkan loader present.

**RCC's own log names the device.** With `RUST_LOG=rog_platform=trace`:

```
[INFO ] Found dgpu 10DE:2520 at "0000:01:00.0"
[INFO ] Found additional device 10DE:228E at "0000:01:00.1"
[TRACE] Looking at PCI device "0000:01:00.0"
[TRACE] Looking at PCI device "0000:01:00.1"      <- last device of every scan; 05:00.0 never reached
```

(Minor: the `trace!("Matched dGPU {id} ... by checking display connections")` line is also printed
for `10DE:228E`, because it is logged before `is_display_class()` decides. Harmless, but misleading
when reading logs.)

**The iGPU fields report the dGPU.** Under a CPU/APU load, so that the two sensors separate:

```
real integrated GPU (amdgpu hwmon, 0000:05:00.0)   62 °C
dGPU (nvidia-smi)                                  50 °C
GUI "iGPU Status | AMD Radeon 680M"                50 °C   <- follows the dGPU
GUI "dGPU Status | GA106M [RTX 3060 Mobile]"       50 °C
```

Idle, both GUI fields read exactly the same number at every sample.

## Expected behaviour

- Opening the GUI should not resume a sleeping dGPU. Any NVML entry point should be gated on the
  dGPU's `runtime_status`, whichever device object happens to reach it.
- The iGPU temperature/usage fields should report the integrated GPU, or nothing — never the dGPU's
  NVML readings.

## Suggested fix

1. **Do not admit non-display functions to the device list.** Add the class check to the sibling
   branch at line 476, e.g. `if dgpu || (!parent.is_empty() && sysname.contains(&parent) && is_display_class(&class))`.
   The audio function has no hwmon, no DRM node and no telemetry to contribute.
2. **Gate the NVML fallback on the vendor test that reaches it, not on `is_dgpu`.** Wherever
   `pci_id` starting with `10DE` is enough to call NVML (lines 315 and 372), the dGPU's
   `runtime_status` should be the thing that decides — `get_gpu_power_status() == GfxPower::Active`,
   a passive sysfs read that (as measured in #293) does not wake the device. That keeps the guard
   correct even if some other non-dGPU NVIDIA function shows up on another board.
3. **Reconsider the `break` at line 492.** It assumes the iGPU is enumerated before the dGPU. On
   boards where the dGPU comes first, enumeration stops after the dGPU's own functions and the iGPU
   is never found, which is what leaves the iGPU telemetry fields to be filled by whatever the
   fallback returns.

## System details

- Distro: Arch Linux
- Kernel: 7.1.8-zen1-3-zen
- Desktop: Hyprland (Wayland)
- Model: ROG Strix G513RM (board G513RM)
- dGPU: NVIDIA GA106M [GeForce RTX 3060 Mobile / Max-Q], PCI `0000:01:00.0` (audio function `0000:01:00.1`)
- iGPU: AMD Radeon 680M, PCI `0000:05:00.0`
- NVIDIA driver: 610.57.04 (open kernel module)
- `/etc/modprobe.d`: `options nvidia NVreg_EnableS0ixPowerManagement=1 NVreg_DynamicPowerManagement=0x02`
- asusctl / rog-control-center: 6.4.0.r0.ge6c1469c-1 (from the `[ogc]` repo)

---

The investigation, the measurements above and this report were produced with the assistance of
Claude (Anthropic's Claude Code). Every figure was measured on the system described above; nothing
is inferred or estimated.

scardracs:
# Description

This PR resolves an issue where discrete GPUs (dGPU) were repeatedly woken up from runtime power management sleep (D3cold) every 2 seconds during ROG Control Center telemetry polling.

### Root Cause & Problem:

1. **Non-display PCI Functions Matching:** Sibling PCI functions in the same PCI slot (such as the NVIDIA High Definition Audio Controller `0000:01:00.1`) were matched in `Device::find()` with `is_dgpu: false`. This bypassed the suspend guard in `probe_hwmon` and `probe_usage`, causing NVML fallback queries to wake the dGPU from suspended state.
2. **Premature Loop Break:** Legacy parent tracking in `Device::find()` included an early `break;` that prevented scanning subsequent PCI buses, potentially missing iGPUs on higher bus indices.

### Key Changes:

- **Display Controller Filtering:** Restricted udev PCI device enumeration strictly to display controller classes (`is_display_class(&class)`), ignoring companion audio or USB-C controllers.
- **NVML Active State Gating:** NVML queries in `probe_hwmon` and `probe_usage` now explicitly verify that the device runtime power status is `GfxPower::Active` before initiating NVML calls.
- **Full Bus Enumeration:** Removed parent prefix tracking and early `break;` in `Device::find()` to ensure all display devices across all PCI buses are detected.
- **Modern Rust & Edition 2024 Optimizations:**
  - Adopted let-chains (`if let ... && let Some(...)`) in `get_gpu_power_status()`, `probe_usage()`, and `Device::find()`.
  - Replaced per-device heap `String` allocation in display detection with zero-alloc `Iterator::any`.
  - Simplified DRM connector parsing in `find_connected_displays()` with `split_once` and `is_ok_and`.
- **Regression Unit Tests:** Added unit tests for non-display PCI class filtering and non-dGPU telemetry behavior when suspended.

Fixes #302

- **ASUS Laptop Model:** ASUS ROG G614PR
- **Kernel Version:** 7.2.0-1-cachyos

# Verification and testing:

### DIFF
--- a/rog-platform/src/gpu_pci.rs
+++ b/rog-platform/src/gpu_pci.rs
@@ -278,9 +278,7 @@ impl Device {
         read: fn(&Path) -> Option<f32>,
         nvml: fn() -> Option<f32>,
     ) -> Option<f32> {
-        if self.is_dgpu
-            && self.get_runtime_status().unwrap_or(GfxPower::Unknown) != GfxPower::Active
-        {
+        if self.is_dgpu && self.get_runtime_status().unwrap_or_default() != GfxPower::Active {
             return None;
         }
 
@@ -310,7 +308,9 @@ impl Device {
         }
 
         // 3. Fallback to NVML if this is an NVIDIA device and hwmon is not available
-        if self.pci_id.to_uppercase().starts_with(NVIDIA_PCI_VENDOR) {
+        if self.pci_id.to_uppercase().starts_with(NVIDIA_PCI_VENDOR)
+            && self.get_runtime_status().unwrap_or_default() == GfxPower::Active
+        {
             return nvml();
         }
 
@@ -322,15 +322,14 @@ impl Device {
         self.probe_hwmon(read_hwmon_temp, read_nvml_temp)
     }
 
-    /// Read the GPU utilization percentage (0.0 - 100.0) from sysfs DRM nodes with NVML fallback.
+    /// Probe this device's DRM directories for usage percentage, falling back to
+    /// `nvml` on NVIDIA hardware whose proprietary driver registers no DRM usage node.
     ///
     /// If this is a discrete GPU and it is not in the `Active` power state,
     /// this immediately returns `None` without accessing DRM sysfs or NVML to prevent
     /// waking the PCIe device from runtime PM sleep.
-    pub fn get_usage_pct(&self) -> Option<f32> {
-        if self.is_dgpu
-            && self.get_runtime_status().unwrap_or(GfxPower::Unknown) != GfxPower::Active
-        {
+    fn probe_usage(&self, nvml: fn() -> Option<f32>) -> Option<f32> {
+        if self.is_dgpu && self.get_runtime_status().unwrap_or_default() != GfxPower::Active {
             return None;
         }
 
@@ -366,12 +365,22 @@ impl Device {
 
         // 4. Fallback to NVML if this is an NVIDIA device and DRM busy is not available
         if self.pci_id.to_uppercase().starts_with(NVIDIA_PCI_VENDOR)
-            && let Some(usage) = read_nvml_usage()
+            && self.get_runtime_status().unwrap_or_default() == GfxPower::Active
+            && let Some(usage) = nvml()
         {
             return Some(usage);
         }
 
         None
+    }
+
+    /// Read the GPU utilization percentage (0.0 - 100.0) from sysfs DRM nodes with NVML fallback.
+    ///
+    /// If this is a discrete GPU and it is not in the `Active` power state,
+    /// this immediately returns `None` without accessing DRM sysfs or NVML to prevent
+    /// waking the PCIe device from runtime PM sleep.
+    pub fn get_usage_pct(&self) -> Option<f32> {
+        self.probe_usage(read_nvml_usage)
     }
 
     /// Read the current graphics clock (MHz) of this GPU from sysfs hwmon with
@@ -383,7 +392,6 @@ impl Device {
     /// Enumerate PCI GPU devices via udev and identify the dGPU.
     pub fn find() -> Result<Vec<Self>> {
         let mut devices = Vec::new();
-        let mut parent = String::new();
 
         let mut enumerator = udev::Enumerator::new().map_err(|err| {
             warn!("{}", err);
@@ -394,14 +402,6 @@ impl Device {
             warn!("{}", err);
             PlatformError::Udev("match_subsystem failed".into(), err)
         })?;
-
-        let get_parent = |dev: &udev::Device| -> String {
-            dev.sysname()
-                .to_string_lossy()
-                .trim_end_matches(char::is_numeric)
-                .trim_end_matches('.')
-                .to_string()
-        };
 
         for device in enumerator.scan_devices().map_err(|err| {
             warn!("{}", err);
@@ -414,18 +414,19 @@ impl Device {
             {
                 let id = id.to_string_lossy();
                 let class = class.to_string_lossy();
-                // Match only Nvidia or AMD
-                if is_gpu_vendor(&id) {
+                // Match only Nvidia or AMD display devices
+                if is_gpu_vendor(&id) && is_display_class(&class) {
                     let mut dgpu = false;
                     // Check connected displays to distinguish dGPU from iGPU.
                     // eDP-1 is the internal panel, always on iGPU.
-                    let displays = find_connected_displays(device.syspath()).unwrap_or_default();
+                    let displays =
+                        find_connected_displays(device.syspath()).unwrap_or_default();
                     if !displays.contains(&"eDP-1".to_string()) {
                         trace!(
                             "Matched dGPU {id} at {:?} by checking display connections",
                             device.sysname()
                         );
-                        dgpu = is_display_class(&class);
+                        dgpu = true;
                     } else {
                         trace!(
                             "Device {id} at {:?} appears to be the iGPU",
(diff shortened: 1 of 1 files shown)


Ghoul4500 (approved):

rog-platform/src/gpu_pci.rs
```diff
@@ -366,14 +365,24 @@ impl Device {
...
     }
 
+    /// Read the GPU utilization percentage (0.0 - 100.0) from sysfs DRM nodes with NVML fallback.
+    ///
+    /// If this is a discrete GPU and it is not in the `Active` power state,
+    /// this immediately returns `None` without accessing DRM sysfs or NVML to prevent
+    /// waking the PCIe device from runtime PM sleep.
+    pub fn get_usage_pct(&self) -> Option<f32> {
+        self.probe_usage(read_nvml_usage)
+    }
```
Ghoul4500, line 384: why do we need this thin wrapper around self.probe_usage. why not call it directly? It looks like self.probe_usage was originally get_usage_pct so this was added in this PR

Ghoul4500 (changes requested):

Ghoul4500:
@coderabbitai full review

rog-platform/src/gpu_pci.rs
scardracs: For 3 different reasons:

1. Clean Public API: get_usage_pct exposes a zero-argument method for external callers so they don't have to know about or pass read_nvml_usage.
2. Dependency Injection for Unit Tests: probe_usage accepts nvml: fn() -> Option<f32>, allowing unit tests to inject a mock/panicking function (panic_on_nvml) to verify that suspended devices never touch NVML without making live C FFI calls and without the risk to wake up the dGPU.
3. Consistency: Mirrors the existing private engine / public wrapper pattern used by probe_hwmon for get_temp and get_freq_mhz.

Ghoul4500 (approved):

PR #374, opened 2026-09-10
fix(rog-aura): fix RGB .ron for various G615

Pelochus:
# Description

My previous contribution missed per-key RGB for G615LW. Along the way, I've made minor fixes to other G615 variants that were incorrectly labeled.

- **ASUS Laptop Model:** G615LW (ROG Strix G16 2025)
- **Linux Distribution:** Bazzite 44 (Fedora 44)
- **Kernel Version:** 7.2.3

# Verification and testing:


Ghoul4500 (changes requested):
If all G615 variants have the exact same capabilties, they can be collapsed into one entry. the reason we haven't done so is that we can't really be certain of that.

Also in this PR itself, the modified records aren't similar to one another (for example basic zones being an empty array on one while another has 4 keys). If you can verify these, feel free to correct them as well

Ghoul4500 (approved):

PR #381, opened 2026-09-15
fix(rog-aura): honour shutdown state for TUF keyboard power states

voidvore:
# Description

`AuraPowerState.shutdown` is silently dropped for TUF laptops: `tuf_to_bytes()` hardcodes the fifth `kbd_rgb_state` byte to `1`, so asusd always writes the shutdown LED state as enabled and whatever the user configured is ignored. Since that write carries the save bit, the forced value is stored in non-volatile EC state on every daemon start and the configuration can never win.

The fifth byte of the TUF interface gates the keyboard LED during the S5 / power-off sequence. Measured on a TUF Gaming A15 FA507NV (BIOS 318, writing to `/sys/class/leds/asus::kbd_backlight/kbd_rgb_state`):

- `1 0 1 0 1` -> arg0 `0x008804BD` for `DEVS(0x00100057)`: the EC lights the keyboard with its power-on default colour (white) while power drops.
- `1 0 1 0 0` -> arg0 `0x000804BD`: the LED stays dark through the power-off sequence.

Only bit 7 of the state flags differs between the two writes. The same bit position is named `shutdown` in `AuraPowerState::new_to_byte()` for the 2021+ keyboards, `rog-control-center` already exposes the matching toggle, and `asusctl aura power-tuf` already sends `shutdown: false` on the D-Bus call - only the TUF byte packing threw the value away. Pass the configured value through instead of forcing it on.

Machines that never touched the setting keep the previous behaviour: `AuraPowerState::default()` has `shutdown` enabled.

- **ASUS Laptop Model:** ASUS TUF Gaming A15 FA507NV
- **Kernel Version:** 7.2.4-3-cachyos

# Verification and testing:

Additional hardware checks performed on the FA507NV:

- Unit test `check_tuf_control_bytes` covers both `shutdown: false` -> `[1, 0, 1, 0, 0]` and `shutdown: true` -> `[1, 0, 1, 0, 1]` for `LaptopKeyboardTuf`.
- A/B on the live EC via the sysfs attribute: with the shutdown bit clear the keyboard stays dark through `poweroff`; with it set the EC runs its white power-off animation. `AuraPowerState::default()` keeps the bit set, so the new code path is byte-identical for anyone who did not change the setting.


Ghoul4500:
This may need some changes to the asusctl cli as well? Can you pls check that. In the meantime I'll see if I can get a friend with an older TUF to test this patch. You can write here any specific paths you want anyone seeing this to test

voidvore:
> This may need some changes to the asusctl cli as well? Can you pls check that. In the meantime I'll see if I can get a friend with an older TUF to test this patch. You can write here any specific paths you want anyone seeing this to test

```sh
echo "1 0 1 0 0" | sudo tee /sys/class/leds/asus::kbd_backlight/kbd_rgb_state
```

Before entering this laptop, when you shut down the computer: if the LED is off, it turns on and a white LED lights up. If the LED is on but in a color other than white (e.g. red), it slowly fades out and then turns on (this doesn't happen on Windows). When you turn the laptop on, it's as if the keyboard backlight calibrates itself, lighting up all colors in 3 different intensity ranges. After the calibration finishes, it switches to the settings you previously configured. Although I couldn't figure out the exact cause of the issue, setting this to 0 (the 5th flag) fixes the problem. The same parameter exists for ROG devices (a bit below the part I changed) but Luke didn't add it for TUFs (he had said by guessing that it should be `keyboard` in a commit message, so I **THINK** it was added based on a guess)


sorry for the PR description, I was in a hurry, it was entirely written by an llm

voidvore:
To test this, you'll need to rebuild asusd and use the patched one instead of the original asusd. So far I've only tested it with simple commands. Since it's hardcoded as 1, even if you toggle the shutdown parameter from the cli or rogcc, it doesn't take effect.

Rashnan:
TLDR: works

Machine info:
ASUS Laptop Model: ASUS TUF Gaming A15 FA506QM
Linux Distribution: CachyOS
Kernel Version: 7.2.4-3-cachyos

Asusctl 6.4:
Used static pattern red.
Shutdown -> replicated the issue. On power off, at end of sequence getting white keyboard light. On power on, getting white -> rainbow -> red

Asusctl 6.5 master:
Same as 6.4. Used rog-control-center to toggle keyboard additional settings boot and shutdown. specifically turned it on and then off. Left it at off, but still got same as 6.4

Asus 6.5 with pr:
Same procedure as 6.5. keyboard turns off properly during shutdown and boot and no white / rainbow light.

Other issue but likely unrelated:
On login screen powers doesn't power on and needs keyboard spam. Unrelated to this though likely related to cachyos update

voidvore:
> Other issue but likely unrelated: On login screen powers doesn't power on and needs keyboard spam. Unrelated to this though likely related to cachyos update

@Rashnan  I think I'm having the same issue. Did this happen when you first turn on the laptop, or after waking from sleep (opening the laptop lid)? To avoid cluttering the PR, could I ask you to write in the [Discord server](https://discord.com/invite/B8GftRW2Hd) or open an issue on GitHub for this?

Rashnan:
> I think I'm having the same issue. Did this happen when you first turn on the laptop, or after waking from sleep (opening the laptop lid)? To avoid cluttering the PR, could I ask you to write in the [Discord server](https://discord.com/invite/B8GftRW2Hd) or open an issue on GitHub for this?

When first turn on (after opening after every shutdown) and not on wake from sleep. 
Alright I have written in the discord server [here](https://discord.com/channels/725125934759411753/1549803472222691378).

Ghoul4500:
> This may need some changes to the asusctl cli as well? Can you pls check that. In the meantime I'll see if I can get a friend with an older TUF to test this patch. You can write here any specific paths you want anyone seeing this to test

@voidvore have u checked that the cli patch has the relevant changes brought? There should be a flag for this bit

